PR #1, opened 2015-10-15
Examples from 5 and 6

osehgol:
These are code examples from videos 5 and 6.


shiffman:
Fantastic, thank you so much!!!

PR #16, opened 2015-11-10
code for videos 8.2 and 8.3

osehgol:
Also documented (in comments), your singing attempt:
Shiffman tries to sing @ 11:24


shiffman:
Wonderful, thank you!!!

PR #17, opened 2015-11-11
update sketch.js

rscharfer:
There was a bug in line 23 and the circles weren't blinking while intersecting.. I fixed it.


shiffman:
Great catch, thank you!

rscharfer:
you bet!

PR #29, opened 2016-06-03
comments and "width" correction

buseca:
- some explanatory comments added, trying to support the learning.
- changed some “width” to “width/2” to show the whole 800 stars in the
canvas. (with the “width” value, some of theme were placed outside of the
real canvas).


shiffman:
wow, this is amazing and SUPER HELPFUL! I keep meaning to go back and write comments in these but haven't made the time.

PR #37, opened 2016-08-18
Bug in the distance checking

kiuny:
There was a minor bug in the `checkStuck` function.

(a+b)^2 = a^2 + b^2 + 2ab


shiffman:
oh whoops!  I should try to remember to cover this on a future episode.

PR #43, opened 2016-11-01
Update mapping color on sketch.js

keebOo:
On mapping color line hu = map(hu, -1, 1,0, 360); probably was  hu = map(hu, -1, 1,0, 255);
So, changing only a little bit I add a variable colorRange to play with.

### DIFF
--- a/challenges/CC_30_phyllotaxis_p5.js/sketch.js
+++ b/challenges/CC_30_phyllotaxis_p5.js/sketch.js
@@ -9,6 +9,7 @@ var c = 3;
 var points = [];
 
 var start = 0;
+var colorRange = [0,255];
 
 function setup() {
   createCanvas(400, 400);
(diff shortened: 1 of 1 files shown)


challenges/CC_30_phyllotaxis_p5.js/sketch.js
```diff
@@ -26,7 +27,7 @@ function draw() {
     var x = r * cos(a);
     var y = r * sin(a);
     var hu = sin(start + i * 0.5);
-    hu = map(hu, -1, 1, 0, 360);
+    hu = map(hu, -1, 1, colorRange[0], colorRange[1]);
```
shiffman: See the p5.js reference: "Setting colorMode(HSB) lets you use the HSB system instead. By default, this is colorMode(HSB, 360, 100, 100, 1). You can also use HSL."

shiffman: Let me know if I'm missing something, that you for the effort you made to do this pull request!

PR #47, opened 2016-12-07
Added Session 1

blinkityblank:
Added session 1 folder with all the related code.


shiffman:
This looks good but maybe the coding challenges should just stay in the coding challenge folder?  Rather than be duplicated.  We could put a `readme.txt` file that points people to it if that's helpful.  What do you think?

blinkityblank:
Yes, that's a good way to proceed

shiffman:
Ok, I'll merge this and we can remove the duplicated files later!

PR #51, opened 2016-12-15
Fixed missing list

sadmoody:
Array was never declared. This breaks the code.


shiffman:
Oh, thank you!  I think the line `var cells = []` should be in `sketch.js` instead though.  Can you update the pull request?  Thank you thank you!

sadmoody:
Done :) Keep up the good work. Wish I had someone like you when I was starting out!

shiffman:
wonderful, thank you!

PR #52, opened 2016-12-17
Fixed errors with linked JS files and made the code match the one presented in the video and CodePen

Naahuel:
Hello! I noticed the code here didn't match the demo you made in CodePen and the video, so I took the liberty of fixing that.

I commented out the reference of rocket.js because I didn't remove the file. I was unsure why that was there and why it was so different. I'm sure it's not in use though.


shiffman:
This is great, thank you!  I actually meant (I think) for this version to have the rocket code in a separate file (`rocket.js`) but I must have missed copy/pasting it over.  Do you think you could add that to this pull request?

Naahuel:
Done! :) 

I simply replaced the Rocket constructor in `rocket.js` because the old one was very different and I think it would be unfamiliar for the people who watched the video and want to examine the code (like me). Also it seemed that I would have to modify the other files quite a lot to make that constructor work.

Hopefully this is OK. Keep doing awesome work!

shiffman:
This is so amazingly helpful, I have a lot of trouble keeping the code up-to-date and matching with the videos so I so appreciate this!!

Naahuel:
My pleasure! I really admire your work and this tutorial taught me a lot. I'll contribute with as much as I can.

PR #54, opened 2017-01-07
Workaround for TRIANGLE_STRIP Issues in sketch.js

PerryTaga:
Draws the geometry using lines instead of boxes. Code has been commented to reflect this.

Line coloring doesn't seem to be working with WebGL canvases at the moment.
Performance testing has had various results on different machines. The generated graphics are much closer to the original tutorial, though.


SacrificeProductions:
Have addressed TRIANGLE_STRIP here https://github.com/CodingTrain/Rainbow-Code/pull/69

shiffman:
Thank you @PerryTaga!  Indeed this is now addressed by #69.

PR #66, opened 2017-02-10
Adding reduce denominator function for the loop

starak:
Removing decimal from input slider and boosted the range
Adding reduce denominator function for the loop
Only trigger draw when input data is changed

### DIFF
--- a/challenges/CC_55_Roses_p5/sketch.js
+++ b/challenges/CC_55_Roses_p5/sketch.js
@@ -11,8 +11,10 @@ var sliderN;
 
 function setup() {
   createCanvas(400, 400);
-  sliderD = createSlider(1, 10, 5, 0.1);
-  sliderN = createSlider(1, 10, 5, 0.1);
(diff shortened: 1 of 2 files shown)


challenges/CC_55_Roses_p5/sketch.js
```diff
@@ -11,28 +11,38 @@ var sliderN;
 
 function setup() {
   createCanvas(400, 400);
-  sliderD = createSlider(1, 10, 5, 0.1);
-  sliderN = createSlider(1, 10, 5, 0.1);
+  sliderD = createSlider(1, 20, 10, 1);
+  sliderN = createSlider(1, 20, 10, 1);
+  sliderD.elt.addEventListener("input",draw);
```
shiffman: Can you use the p5.dom `input()` function?  `sliderD.input(draw)`

shiffman:
Thank you, such nice improvements!

PR #69, opened 2017-02-18
Update sketch.js

SacrificeProductions:
Fix for using TRIANGLE_STRIP not working properly with p5, the need for specify a fill and using opacity to help visual aid as have not looked into shading.

PR #70, opened 2017-02-18
Create sketch-v2.js

SacrificeProductions:
Add variation to the original code without changing the master challenge code.


shiffman:
This is great!  I think you can actually just put the edits directly in sketch.js, no need to keep the boxes.  Also, I think it's best if you pull request a version without your clever color variations.  Best to keep the example as simple as possible then others can create their own design variations.

shiffman:
I think this is actually covered in #69!  Since this version has your own creative spin I think it's best for it to live in your own repo rather than in the examples?  Not sure the best way to handle this.

SacrificeProductions:
Still learning github lol, will work something out. Whoops

shiffman:
You're doing great!

PR #75, opened 2017-02-22
Added way to change words

tslytsly:
added a simple function to change the text when space is pressed.
It also checks if there are more or less vehicles and pushes or pops the array.

### DIFF
--- a/challenges/CC_59_Steering_Text_Paths/sketch.js
+++ b/challenges/CC_59_Steering_Text_Paths/sketch.js
@@ -5,6 +5,9 @@
 
 var font;
 var vehicles = [];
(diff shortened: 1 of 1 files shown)


shiffman:
Hey this is fantastic!  This repo, however, is meant to have the code that matches the video exactly so others can start from that code and make their own improvements.  I was thinking of putting READMEs in the challenge folders then people could link out to their improved code.  Would you like to pull request that?

tslytsly:
@shiffman yea sure.

I'm not entirely sure how to do that. Don't use github much,

Love the videos BTW, great teaching of concepts that apply to lots of Object oriented languages.

PR #77, opened 2017-02-22
merge drawing steps into one loop

etothepi16:
Per one of my earlier pull requests, resubmitted because I went off scope on that one.
The new loop requires changing da to keep the resolution the same.

### DIFF
--- a/challenges/CC_60_Butterfly_Wings/sketch.js
+++ b/challenges/CC_60_Butterfly_Wings/sketch.js
@@ -18,22 +18,12 @@ function draw() {
   fill(255, 50);
   strokeWeight(1);
 
-  var da = PI / 100;
+  var da = PI / 200;
   var dx = 0.05;
 
   var xoff = 0;
   beginShape();
-  for (var a = -PI / 2; a <= PI / 2; a += da) {
-    var n = noise(xoff, yoff);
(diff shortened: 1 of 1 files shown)


challenges/CC_60_Butterfly_Wings/sketch.js
```diff
@@ -18,22 +18,12 @@ function draw() {
...
-    var r = sin(2 * a) * map(n, 0, 1, 50, 300);
-    var x = r * cos(a);
-    var y = r * sin(a);
-    xoff += dx;
-    //point(x, y);
-    vertex(x, y);
-  }
-
-  for (var a = PI / 2; a <= 3 * PI / 2; a += da) {
+  for (var a = -PI / 2; a <= 3 * PI / 2; a += da) {
```
shiffman: I didn't test this cod but doesn't there need to be something to increment xoff for half the loop and decrement it for the other half?

etothepi16: Yes, if you want it to have symmetry. Personally, I like the asymmetrical look. You can get symmetry back by adding `xoff %= 5;`

etothepi16: True. fixed

shiffman:
nice, thank you!!

PR #83, opened 2017-02-24
Update README.md

tslytsly:
Added link for running version, as per @shiffman 's suggestion. Moved source code to another link.


shiffman:
Yay!

PR #86, opened 2017-02-24
Add community variation

JosefKuchar:
Hello,
can you add my community low poly variation?


shiffman:
This is fantastic. . and I showed it today on the live stream!

JosefKuchar:
I know, i was on the livestream as pepa489 in the chat

PR #87, opened 2017-02-24
simplified code with ternary operator

etothepi16:


shiffman:
I agree this is a nice improvement!  However, I tend to avoid using ternary operators in my examples to be friendlier to beginners.

PR #89, opened 2017-02-27
Move the Star object to a new js file

MarkovicNikola:
Move the star object to a new js file and updated sketch.js and index.html


shiffman:
Thank you!

PR #92, opened 2017-02-27
Update sketch.js

joshgreensmith:
Just to adjust the for loop to go from 0 to two_pi to make it more intuitive


shiffman:
Thanks!

PR #93, opened 2017-02-27
Added variant to the README

Gaweph:
Just finished my variant, was hoping to add to the list.  Words morph into one another and added a splash of colour :)
Also source code was migrated to typescript (ES6).


shiffman:
Yay, thank you!

PR #96, opened 2017-02-28
Add rainbow fractal in p5.js

Zalastax:
I'm not completely happy with https://github.com/Zalastax/p5js-projects/blob/master/CC_61_fractal_spirograph_rainbow/sketch.js#L51-L52 though, because of how bad the performance gets. Any way of creating performant gradient line shapes in p5.js?


shiffman:
Lovely!

PR #99, opened 2017-03-03
Add links, clean some code

MarkovicNikola:
Added Patreon and video links and removed some comments that stayed in the files but weren't very useful


challenges/CC_61_fractal_spirograph/CC_61_fractal_spirograph.pde
```diff
@@ -10,12 +11,13 @@ int resolution = 50;
...
 Orbit sun;
 Orbit end;
+Orbit next;
 
 void setup() {
   size(600, 600);
   path = new ArrayList<PVector>();
   sun = new Orbit(width/2, height/2, width/4, 0);
-  Orbit next = sun;
+  next = sun;
```
shiffman: I think I prefer to leave next as a local variable for the loop only? It's comparable to `int i = 0` in a for loop. Is that ok with you?

MarkovicNikola: Yes, of course, I just thought it made sense to make it a global variable instead of declaring it every time since we were using it in three different places, However, the `int i = 0` comparison is on point, I agree and have reverted the changes

shiffman:
Wonderful! Would you also like to add a link to the video in the comments at the top?

https://youtu.be/0dwJ-bkJwDI

MarkovicNikola:
Of course Daniel :smile:

shiffman:
Yay, thank you for this help!!

PR #106, opened 2017-03-08
Add links, consistent spacing

MarkovicNikola:
Just updated this coding challenge with some consistent spacing and general clean up
I also added links to the site, patreon and video


shiffman:
Thank you for this, so helpful!

PR #108, opened 2017-03-09
Add links, clean the code

MarkovicNikola:
Added the links and made code a bit easier to read, also, I moved the gravity variable to the CC_31_FlappyBird.pde since it was already global. I think it's best to have all of the globals in a single place, even if they are only used in one file, I'll be happy to revert the change if you don't agree :)


shiffman:
wonderful thank you for this!!

PR #110, opened 2017-03-12
Added my variant to the README

MCManuelLP:
I know I'm a bit late but it just took a bit longer than planned.
I added a choice of fonts, vehicles 'split' and 'join' to accommodate for the needs of the new string and you can choose any colours you want.


shiffman:
Fantastic, thank you!!

PR #114, opened 2017-03-14
Better locking of the flag

The-Code-Beetle:
I locked the whole left column of the flag, so it will connected to the pole.
`for(int i = 0; i < particles[0].length; i += 1){
    particles[0][i].lock();
}`


shiffman:
Oh, I like this change!

PR #117, opened 2017-03-17
Bug: selfAngle not initialised for first segment.

heemonsu:
Since the selfAngle variable is not initialised for the segment with no parent, the animations always begin from angle = 0. If we look at the animation around 1:25:40 in the live stream video, the segments begin to rotate from angle = 0, even though the first segment is defined to start from angle = -45.

### DIFF
--- a/challenges/CC_64_1_a_ForwardKinematics/Segment.pde
+++ b/challenges/CC_64_1_a_ForwardKinematics/Segment.pde
@@ -26,6 +26,7 @@ class Segment {
     a = new PVector(x, y);
     len = len_;
     angle = angle_;
+    selfAngle = angle;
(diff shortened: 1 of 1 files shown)


shiffman:
Ah, thank you!!

heemonsu:
BTW, I didn't know processing initializes uninitialized variables to the default value. Kind of an unfavourable thing to have, in my opinion. What are your thoughts?

shiffman:
This is actually the default behavior of Java, so it's not Processing and not something that could really be changed. But yes, it is sometimes convenient and sometimes causes confusion!

heemonsu:
Thank you for clarifying, Daniel!

PR #118, opened 2017-03-17
Plinko README.md

storealutes:
Created a readme doc for CC 62 Plinko. This has a table of links to other repos of this same project.


shiffman:
Thank you!

PR #132, opened 2017-03-31
Updated readme with link to ruby version

jwood74:
I made a ruby version of the Breadth-First-Search code. Thought it was worth sharing


shiffman:
Awesome, thank you!

PR #134, opened 2017-04-03
fixed  adding item into  tail[-1] when total = 0

himaytoi:


shiffman:
Thanks!

himaytoi:
function setup() {
   console.log("Great videos!");
}

function draw() {
   console.log("Thank you!!!");
}

PR #148, opened 2017-05-01
Update to Follow Bot

barbarosyurttagul:
In this project, if the user follows someone, the follow bot works again and it tweets like '.@yourTwitterName do you like rainbows'.
So, to prevent this, we make an if clause.

### DIFF
--- a/bots/node4_b_followbot/bot.js
+++ b/bots/node4_b_followbot/bot.js
@@ -15,7 +15,14 @@ function followed(eventMsg) {
   console.log("Follow event!");
   var name = eventMsg.source.name;
   var screenName = eventMsg.source.screen_name;
-  tweetIt('.@' + screenName + ' do you like rainbows?');
+  /* If you follow someone, follow bot works again.
+     It tweets like '.@yourTwitterName do you like rainbows'
+     So, to prevent this, we make an if clause
+     id_str is your twitter account id (eventMsg.source.id_str)
+     */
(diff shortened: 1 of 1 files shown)


bots/node4_b_followbot/bot.js
```diff
@@ -15,7 +15,14 @@ function followed(eventMsg) {
   console.log("Follow event!");
   var name = eventMsg.source.name;
   var screenName = eventMsg.source.screen_name;
-  tweetIt('.@' + screenName + ' do you like rainbows?');
+  /* If you follow someone, follow bot works again.
+     It tweets like '.@yourTwitterName do you like rainbows'
+     So, to prevent this, we make an if clause
+     id_str is your twitter account id (eventMsg.source.id_str)
+     */
+  if(eventMsg.source.id_str !== 'your_twitter_account_id_here'){
```
shiffman: Is there a way to do this with your twitter account name rather than your id string?

shiffman: (Thank you for this btw!!!)

barbarosyurttagul: Yes of course we can do this by screen name like
if(eventMsg.source.screen_name !== 'yourScreenName'). 
Btw if you want like this I can resend pull request.

shiffman: Yes, I think that's easier for people to understand, rather than know how to look up their twitter id? If you can resubmit that would be great!

meiamsome:
Closing in preference of #149

PR #149, opened 2017-05-01
Update to Follow Bot

barbarosyurttagul:
If you follow someone, follow bot works again. It tweets like '.@yourTwitterName do you like rainbows'. So, to prevent this, we make an if clause using your twitter screenName.


meiamsome:
Thanks for this, it's now merged (Took long enough!)

PR #151, opened 2017-05-03
Challenge 35 (TSP) - Ported to TypeScript

Gaweph:
I wasn't sure where to put this.
I created a folder in TSP 35: _challenges/CC_35_TSP/CC_35.4_TSP_GA_TypeScript/_
Let me know if this is not correct.

@shiffman, I moved some of the logic into 2 classes **Population** and **PopulationItem**.  Would you prefer a straight port to typescript in future (preserving your exact code layout)?


shiffman:
Awesome! I wonder if this would make more sense in its own repo and then linked outwards from the README file in the TSP folder?

Gaweph:
@shiffman, No problemo.  I will move to the readme.  Which readme?  Its not a variant as such.  I could just scrap this one, and then I will make a variant using typescript and just link that instead.

shiffman:
A port to another language / environment absolutely counts as a variant!  Feel free to add to this README or wherever you think is most appropriate.

https://github.com/CodingTrain/Rainbow-Code/tree/master/CodingChallenges/CC_35_TSP/CC_35.4_TSP_GA

PR #154, opened 2017-05-07
Updated README and updated file tree

nobobo1234:
Reply if i accedientelly missed or deleted something. Willing to make a site, wait for that


shiffman:
This looks great! I'm going to merge it in then make some small tweaks ("challenges" is spelled incorrectly for example.) I'm also noticing that some of my examples don't match the YouTube numbering so would be great to fix these up sometime soon. Thank you for this help!!

PR #155, opened 2017-05-11
Update box.js

TusharSadhwani:
Box generate function was faulty.


shiffman:
Can you explain what this change is for? Thanks!

TusharSadhwani:
The new boxes after each generation were at a distance (= their size) from all other boxes. They should have been stacked together. The menger sponge was getting gaps in it.

I know, I'm really bad at explaining. The code wasn't working before, it works now :P

shiffman:
Wonderful, thank you! And your explanation is perfect!

PR #157, opened 2017-05-12
Added Port to typescript

Gaweph:
@shiffman As discussed, here you go :)


Gaweph:
This request is relating to the discussion here: #151

PR #158, opened 2017-05-12
Added some documentation for functions

penguoir:
Added some documentation for functions using standard [JSDoc](http://usejsdoc.org/)

### DIFF
--- a/CodingChallenges/CC_71_minesweeper/index.html
+++ b/CodingChallenges/CC_71_minesweeper/index.html
@@ -1,12 +1,8 @@
 <!DOCTYPE html>
 <html>
-  <head>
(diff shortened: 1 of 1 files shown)


shiffman:
This is great! I'm wondering, however, if the formal nature of the JSDocs syntax is confusing/overwhelming for a beginner audience? Would more explanatory comments be better? I do not have a strong opinion one way or another, curious to hear all points of view. Thank you!

Gaweph:
@penguoir I think perhaps some of the changes to the code are incorrect, for example the **make2DArray** method does not create a 2d array anymore.

meiamsome:
I'm going to close this and not merge based on the previous two commits - It's great to comment it with standard comments, but not that great for beginners. Breaking `make2DArray` is also a bit of an unnecessary thing. Hopefully that's okay - I'm happy to discuss if you disagree. Thanks

PR #160, opened 2017-05-14
Extended Beesweeper

kleinesfilmroellchen:
Hi Dan and everybody else on the internet,

This is my version of the Beesweeper coding challenge as it was directly after the livestream. I added base functionality like better feedback, win condition, and the ability to place flags. But, I also gave the page a better design, added images, sound effects and a How To Play. (Sorry for making everything green, that's just my favorite color). I'm looking forward to seeing my contribution in a video soon!

kleines filmröllchen


shiffman:
This is amazing! I like to keep the code here matching what is in the video and then link outwards from a `README.md` file to variations. Can you put your code in another repo then do a pull request to add a README here? You can see some examples in other challenge folders. Thank you!!!

For more (this could probably use some work):
https://github.com/CodingTrain/Rainbow-Code/blob/master/CONTRIBUTING.md

PR #161, opened 2017-05-15
Create README.md

kleinesfilmroellchen:
as asked for by dan
added link to my version

PR #163, opened 2017-05-19
Added variants file for CC_71 minesweeper

Gaweph:
I have also added the previous users variation above mine (his pull request is still open - this should cover both)

@shiffman As noted above I have added the previous users variation which is still an open pull request


shiffman:
Thanks! Merging this one in lieu of #161

Gaweph:
@shiffman looks like I missed a slash in the path, the readme has gone into the root of coding challenges folder.  My apologies

Gaweph:
Created #164 to correct this error

PR #166, opened 2017-05-25
Update cell.js

JAD3N:
Skips rows for columns outside usable range.


shiffman:
This is a nice improvement, thank you.

PR #171, opened 2017-05-27
Minor improvements in tree grow()

lmtuan:
- Line 37: init record to max_dist instead of the abitrary 100000
- Line 45: remove the redundant condition checking
- Line 70: only reset if branch is attracted to some leaves, rather than reset all branches


shiffman:
This is great, thank you!

PR #173, opened 2017-05-28
Declare boxes array within setup to fix array filling error

joeyklee:
It seems there’s an issue in Firefox if the `boxes` array is not
declared within setup. As a result when looping through draw, something goes wonky with the
`boxes` variable and we get `boxes[checkIndex] is
undefined` error. I hope this helps! Thanks for all the great work and materials 🌈❤️!


meiamsome:
Hey there, I just checked and it's working fine on my version of firefox - were you running the example from the repository? It's possible that a different version of p5 could cause this.

joeyklee:
@meiamsome - thanks for looking into this! Actually we can close this PR. With the latest P5.js things are back in order :) hooray!

PR #177, opened 2017-06-12
Added a python implementation

Sayan98:


Zwork101:
I second this, I myself have written many of these programs in python. For example, here is agar.io in prossesing.py (Not online, but I could add that if I wanted) [Gist here](https://gist.github.com/Zwork101/04d363e619d1997e1e37ff6851b83563) I have also created The wandering salesman, snake, and a couple others. So, at least 1 video with prossesing.py would be AMAZING :D

Zwork101:
Although, not sure what the this repo benefits from the file change suggested

PR #179, opened 2017-06-15
Add a list of all coding challenges with a link to his folder

arisanguinetti:
Hi, I've updated the readme.md to have a link between each folder and the corresponding coding challenge so it's easier to track which ones need to be ported from one language to the other. it's still pending to add a link for each coding challenge to each video on YouTube ;)


shiffman:
This is amazing, thank you!

PR #205, opened 2017-09-03
changed clat and clon in the main url

ZaurBilalov:
Hey, I guess you misplaced clat and clon on the URL. The documentation says longitude should come first,
for example:
"https://api.mapbox.com/styles/v1/mapbox/streets-v10/static/-122.4241,37.78,14.25,0,60/600x600?access_token=your-access-token"


meiamsome:
That's a good spot, thanks

PR #215, opened 2017-09-22
p5Dojo

ChristerNilsson:
p5Dojo


meiamsome:
I am a fan of the technical clocks like the QR codes - even if they aren't that readable!

ChristerNilsson:
Yeah, I know. 
Had to construct a neural network to decode it. 
I used Tensorflow. Gets right 99% of the time.

ChristerNilsson:
My only problem is, it is a bit hard to explain missing some meetings.

PR #216, opened 2017-09-22
My Clock

7Flixs:
You can view it here https://alpha.editor.p5js.org/Flixs/sketches/BkaElgmo-, and the soruce is here https://github.com/7Flixs/p5_js_clock

I tried to enhanced your idea.


meiamsome:
I like what you've done here... The use of `lerp` is a nice feature

PR #218, opened 2017-09-22
Add Dave Briccetti’s 3D clock

dcbriccetti:
Atom trimmed some other lines. Sorry for the complication.


meiamsome:
That's pretty nice!

No problem regarding the trimming of other lines!

PR #219, opened 2017-09-23
Update README.md

toferj:
Just adding my little clock... not impressive in the least


meiamsome:
Thanks, that looks good

PR #220, opened 2017-09-23
Update README so my clock is in

147loch:
I made a binary clock, added the link in the readme. But here it is:
[Binary Clock](http://147lo.ch/clock-p5), [Source](http://147lo.ch/clock-p5-source)


meiamsome:
Nice Clock!

PR #227, opened 2017-09-25
Added a 10 PRINT to community editions

Merijn-DH:


meiamsome:
Hah, that's pretty nifty!

PR #233, opened 2017-09-27
laying twigs

madacoo:


meiamsome:
That's a nice alternate way to place the lines and it creates a nice, and different, effect.

PR #241, opened 2017-09-29
Add twitter hashtag for more examples

LiteSoul:
Good idea or not?


shiffman:
Yes, I wonder if thinking of a hashtag for each challenge would be useful? But it certainly works great here!

PR #249, opened 2017-10-03
add 10print 3D example from Michael Hazani

MichaelHazani:
Added a small 10print doodle in A-Frame/threejs to the examples.


meiamsome:
That's pretty cool

MichaelHazani:
Thanks @meiamsome ! Couldn't get the sdf font to behave more "10printy" (i.e make clean lines) in the time that I'd allotted but that's def on my to-do list.

PR #251, opened 2017-10-04
Add CodePen link for Wikipedia Challenge

anku-tiwari:


meiamsome:
This is a good idea, but I'm going to prefer the GitHub pages site rather than codepen, sorry.

PR #264, opened 2017-10-11
Update README.md

20hz20khz:


meiamsome:
Nice, that looks like a pair of cones!

PR #277, opened 2017-10-13
Added my version of the particle challenge

Tr3v0rC:


meiamsome:
This looks nice, love the pastel colours!

Tr3v0rC:
@meiamsome Thanks! I attempted to make them appear as though there were going in all directions, sort of like a sprinkler... Sadly though they really only look like they're coming directly at you. If you look at the smaller particles you will see them getting smaller but the effect didn't work well :-(.

Issue #267, opened 2017-10-11
Have all Coding Challenges in Processing & p5 


meiamsome:
List of challenges:

- [x] Coding Challenge 1: Starfield in Processing
- [x] Coding Challenge 2: Menger Sponge Fractal
- [x] Coding Challenge 3: The Snake Game
- [x] Coding Challenge 4: Purple Rain in Processing
- [x] Coding Challenge 5: Space Invaders
- [x] Coding Challenge 6: Mitosis Simulation
- [x] Coding Challenge 7: Solar System in Processing - Part 1 (2D)
- [ ] Coding Challenge 8: Solar System in Processing - Part 2 (3D)
- [ ] Coding Challenge 9: Solar System in Processing - Part 3 (3D textures)
- [x] Coding Challenge 10: Maze Generator
- [x] Coding Challenge 11: 3D Terrain Generation with Perlin Noise in Processing
- [x] Coding Challenge 12: The Lorenz Attractor
- [x] Coding Challenge 13: Reaction Diffusion Algorithm in p5.js
- [x] Coding Challenge 14: Fractal Trees - Recursive
- [x] Coding Challenge 15: Fractal Trees - Object Oriented
- [x] Coding Challenge 16: Fractal Trees - L-System
- [x] Coding Challenge 17: Fractal Trees - Space Colonization
- [ ] Coding Challenge 18: 3D Fractal Trees
- [x] Coding Challenge 19: Superellipse
- [ ] Coding Challenge 20: 3D Cloth with toxiclibs
- [x] Coding Challenge 21: Mandelbrot Set
- [ ] Coding Challenge 22: Julia Set
- [x] Coding Challenge 23: 2D Supershapes
- [x] Coding Challenge 24: Perlin Noise Flow Field
- [ ] Coding Challenge 25: Spherical Geometry
- [ ] Coding Challenge 26: 3D Supershapes
- [x] Coding Challenge 27: Fireworks!
- [x] Coding Challenge 28: Metaballs
- [x] Coding Challenge 29: Smart Rockets
- [x] Coding Challenge 30: Phyllotaxis
- [x] Coding Challenge 31: Flappy Bird
- [ ] Coding Challenge 32: Agar.io:
- - [x] Part 1
- - [ ] Part 2 (Websocket)
- [x] Coding Challenge 33: Poisson-disc Sampling
- [x] Coding Challenge 34: Diffusion-Limited Aggregation
- [x] Coding Challenge 35: Traveling Salesperson:
[...]

PR #278, opened 2017-10-14
port of CC_67_Pong to p5.js

madacoo:
Pong too! (for issue #267)

I've directly ported Shiffman's pde code to JavaScript so that it should function basically identically. I did notice that there were a lot of spots where refactoring wouldn't hurt, but I thought it would be better to keep the code consistent with how the pde version works so that anyone following along on the coding challenge video would have comparable Javascript code to look at.

Cheers


meiamsome:
Yep, I think we are preferring a direct port over cleaning up code! These ports are lovely, by the way!

(Did you mean to use a different email on your commit in here? It's not linking to your github profile anymore)

shiffman:
yay, nice work, thank you!!

madacoo:
@shiffman You're absolutely welcome!

@meiamsome Oops, no that was not intentional. Thanks for letting me know. Better fix that.

PR #288, opened 2017-10-17
Update README.md

neoxharsh:
Check my version of the clock


meiamsome:
Oooh, that looks nice!

PR #290, opened 2017-10-18
Added guess the number to CC79

olback:


shiffman:
Exciting to see a variation of this! Nice design!

PR #291, opened 2017-10-20
Add Dave Briccetti’s “Stupid Nonsense Spoken Sentences”

dcbriccetti:


meiamsome:
I like it! It looks like I can't run the example online because my chrome won't permit it (I assume because it's not over HTTPS), which was sad - but I had a look at your video and it seems pretty cool.

dcbriccetti:
Chrome won’t let me run it from there either. I was hoping it’s just me. I copied it to another of my domains where I have HTTPS running, and it works:
https://dbschools.com/stupid-nonsense-spoken-sentences

dcbriccetti:
I added a redirect from the first (HTTP) site to the second (HTTPS).

meiamsome:
Ah, that's great!

PR #295, opened 2017-10-25
added code using let and classes

nexusbug:
Updated code using let/classes instead of var/constructor function


shiffman:
This is great thank you! I wonder if there's a better way to have both versions (ES5, ES6) of certain challenges that is more clear than `sketch_2.js`. @meiamsome if you have any brilliant ideas.  A separate subfolder with ES6? A separate higher level directory?

nexusbug:
Thank you, I really appreciate the feedback. I'm sorry, I wasn't very present when I made the pull request with an unclear name like "sketch_2". 

What do you think is a good idea for directory/naming? I have more of those code on the way with updated let and classes, also for Nature of Code too. I'm will make next pull requests according to your preference.

meiamsome:
The folder seems like a good idea.

If there is going to be a website setup, it might make sense to keep code in old version of JS & in ES6 but this is so far off that I think having the ES6 version in the folder is the best bet for the moment. We can always change it later if necessary.

PR #298, opened 2017-10-31
Made links in readme files human readable.

G4m3M4ni4c:


shiffman:
Thank you for this, so so helpful!

PR #300, opened 2017-11-02
bubble.js with es6

mehedi432:
This is updated to es6 & in tutorials there is used es5 . This is updated version of es6 .


mehedi432:
In 7.6 Shiffman uses es5 and i want to update this with es6. I am a learner
& still learning from Shiffman. So, i thought that it is good to share with
what i just learned.

Merijn-DH:
See my comment in pull request #301

PR #301, opened 2017-11-02
using let instead of var

mehedi432:
try to make it es6.


Merijn-DH:
I think we want to keep the code the same as in the tutorial to prevent confusion, but feel free to add an extra folder ES6_Version/ with updated code. (as can be seen here: https://github.com/CodingTrain/Rainbow-Code/tree/master/Tutorials/P5JS/p5.js/09/9.07_p5.js_Drawing_Object_Trails)

PR #302, opened 2017-11-03
bubble.js with es6

mehedi432:
This is made up with es6.


Merijn-DH:
See my comment in pull request #301

PR #304, opened 2017-11-07
Patch 4

mehedi432:
This is for version es6. In tutorial Shiffman uses es5.


shiffman:
Thanks for doing all of these, apologies that these already got uploaded separate from your pull request.

mehedi432:
thank you Sir,

PR #308, opened 2017-11-10
Add past livestreams list

lucakiebel:
This looked like it needed some love. All video titles link to the playlist at their respective position!


lucakiebel:
I will also look into getting this to work automatically once there's a new video in the Live Stream playlist

shiffman:
This is so so helpful and wonderful, thank you!

PR #311, opened 2017-11-18
ES6

mehedi432:
This is made with es6 using class and in new object eoriented way.


mehedi432:
okay....

shiffman:
I think the ES6 version is in the repo already, thank you!!

PR #312, opened 2017-11-18
V_ES6

mehedi432:
With es6 version using class and let variable...


shiffman:
I believe the ES6 code is already there in 7.7.  Feel free to check and see if there any mistakes in my version!

PR #314, opened 2017-11-18
Ported CC 10 to Processing

Merijn-DH:
Issue #267


meiamsome:
Looks great as always!

PR #318, opened 2017-11-22
Added my version of the clock

ignazioc:


addyh:
That's cool! Have you thought about making the colon blink at the same time as the second changes?

ignazioc:
was like this at the beginning, but it was a bit "slow", so now blinks faster. But I also agree with you that's a bit annoying. Let's rename this clock "OCD torture clock" :D

ignazioc:
This is the other version https://alpha.editor.p5js.org/ignazioc/sketches/B1bPjAfef

meiamsome:
Hey, this is great! Just thought I'd point out that the 60 never lights

PR #319, opened 2017-11-23
Relaxing bears extension added

AlcinaW:
v1 (Thank you for making the video series.)


meiamsome:
Those are some cute bears

shiffman:
This is so wonderful.

PR #322, opened 2017-11-29
Fix simetry

JuanIrache:
The condition on line 31 was producing 3 equal wings and one different, instead of two and two. Once that's changed, the rotation is not necessary, it produced simetry on the horizontal axis instead of the vertical one.


shiffman:
This seems like it is more of a "user variation" with adding color and changing the design a bit. Can you add a link to your version in a README instead? If there are small bugs with the existing version as matches in the video that can be pull requested here. 

Thank you for the contribution!!

JuanIrache:
@shiffman Sorry, for some reason my own customisations were included in the pull request, in addition to the fix. I've created a new pr that should just fix the simetry issue.

PR #342, opened 2017-12-26
Filled in blank Past Live Streams

spiray:
In past live streams Live Stream 18-26 did not have a description. I copied each of the title verbatim from YouTube.


shiffman:
So helpful, thank you!

PR #352, opened 2018-01-04
Fix simetry (updated)

JuanIrache:
(Sorry, my previous pull request somehow included my own customisations, in addition to the fix. I hope it's fine now)

The condition on line 31 was producing 3 equal wings and one different, instead of two and two. Once that's changed, the rotation is not necessary, it produced simetry on the horizontal axis instead of the vertical one.


shiffman:
Thank you!

PR #361, opened 2018-01-08
add javascript version from langtonsant

vortami:


vortami:
Whoops, accidently `pixelDensity(5);` is still in there on line 13.

shiffman:
Thank you! Just curious did you try using `pixels` as well or just `point()`?

vortami:
I tried it with pixels, but it didn't work. So I just used `point();`.

vortami:
If I used the grid to draw it, it would run very slow.

PR #368, opened 2018-01-10
Add new community variation to README

omareq:


shiffman:
Wonderful! Also, see this new system:

https://github.com/CodingTrain/Rainbow-Code/wiki/Community-Contributions-Guide
https://github.com/CodingTrain/Rainbow-Code/blob/master/_CodingChallenges/89-langtonsant.md

omareq:
Thanks.  Just to be clear for next time should I only add the contribution to the collection or should I add it to the corresponding readme as well?

shiffman:
I think going forward the README's will go away, but while we're working out the kinks for the new system having them in both is great. Thank you!

PR #369, opened 2018-01-10
Created most of the .md files for the Coding Challenge series

blinkityblank:


shiffman:
Looks like challenge 1 uses underscore `01_starfield.md" whereas the rest use hyphen `-`.  @nilsweb any reason we should go with one over the other?

nilsweb:
Looks nicer? I think it’s kind of an unofficial standard to use hyphenated URLs (or at least I haven’t seen many websites that use underscores).

shiffman:
🎉🎉

PR #378, opened 2018-01-14
Processing port for CC 71, 81.1, 81.2, 85 and 88

Merijn-DH:
Issue #267


shiffman:
Amazing, thank you!!!

PR #379, opened 2018-01-14
Added Processingversion of CC_15_FractalTreeArray #267

lustigo:
Added Processing Version for Fractal Tree Array (Object Oriented)
Used ArrayList instead of Arrays (Java Arrays have a deterministic size, the ArrayLists are more than Javascript Arrays)


shiffman:
Looks great, thank you!

Issue #359, opened 2018-01-08
Challenges with p5.js and Processing code


shiffman:
Some challenges (or in an ideal world perhaps all?) have both p5.js code and Processing code. The Processing code does not run as an online demo. An example is the Langton's Ant. Right now there is only Processing code which is linked from "Get the Code". The "Live Example" points to 404. This is what I'm thinking:

```
repository_p5: /CC_89_langtonsant_p5
repository_Processing: /CC_89_langtonsant
```

"Live Example" would appear only if there was a "repository_p5" link? And download buttons would appear for one or both or p5 / Processing?

Would something like this work?

shiffman:
As discussed on slack, if neither is present, then a message says "code now available yet" would appear . . .perhaps with a link to a wiki on how to contribute it.

nilsweb:
That would probably be a good idea. As for now you can also set `live_example: false` (or to some URL) to disable (or change) the location of the live example.

I think that it should also be possible to detect, if a file already exists using Jekyll. Therefore we could stick with the current `repository` property and let Jekyll automatically figure out what's available so there would be no need to update the challenge's definition file when adding new code.
This would only require a standardised naming scheme like `CC_XX_tile_p5` for p5 and `CC_XX_title_processing` for Processing.

Maybe we could also add a static page that runs the web version of a Processing sketch but I haven't worked with those before so I don't know if this is could be done easily.

shiffman:
Awesome, this sounds great. I don't mind maintaining an extra `live_example` field if that makes it easier than a naming convention / auto-detection. 

It's not worth trying to get web versions of Processing sketches to run, I'm either doing it in Processing b/c it's not appropriate for the browser or if it is, I might as well port to p5.

nilsweb:
The only reason for having an extra `live_example` field is just so you can turn it off (by setting `live_example: false`) or link to an external resource (e. g. a Node.js-based project hosted on Heroku). So it wouldn't make it easier nor more difficult.

I'll make some experiments on auto-detection and report back if it would work or not.

shiffman:
all sounds great! I like the idea of being able to link to an external resource. For some of the videos the example code or demo ends up living in a different repo.

nilsweb:
So I'll add the ability to reference external repositories as well since it is now supported at the moment.

PR #380, opened 2018-01-14
Auto-detection for p5.js and Processing code

nilsweb:
In reference to #359

From the first commit:
- If multiple code variations are found for one repository, they will be
  displayed as a dropdown when downloading the code.
- If one of the found code variations is suitable as a live example, it
  will also be used as that.
- If there is only one variation for the given repository, it will be used
  when downloading the code.
- If no variation is found, the repository will be used "as is" (without
  adding a suffix) for downloading and as the live_example.
- If you link to an external repository, it will be used for downloading
  but not as the live example.
- Specifiying live_example will override any automatically set live example.
- `live_example: false` will disable the live example button.
- Setting `repository: false` will display a message saying that the code
  hasn't been added yet (but will be added in the future).
- Setting `repository: null` (or omitting it) will hide any code actions.The following behaviour is to be expected:
- If multiple code variations are found for one repository, they will be
  displayed as a dropdown when downloading the code.
- If one of the found code variations is suitable as a live example, it
  will also be used as that.
- If there is only one variation for the given repository, it will be used
  when downloading the code.
- If no variation is found, the repository will be used "as is" (without
  adding a suffix) for downloading and as the live_example.
- If you link to an external repository, it will be used for downloading
  but not as the live example.
- Specifiying live_example will override any automatically set live example.
- `live_example: false` will disable the live example button.
- Setting `repository: false` will display a message saying that the code
  hasn't been added yet (but will be added in the future).
- Setting `repository: null` (or omitting it) will hide any code actions.

**The documentation also needs to be updated to reflect the new behaviour!**


shiffman:
This all sounds great! I'm slightly worried it'll be complicated for contributors but let's give it a try and in the end I think it will actually be quite simple once we have a "template" for people to copy from.

Also, one thing to watch out for, if you rename a directory for a Processing sketch, i.e.

```
CC_Snake/ --> CC_Snake_Processing/
```

you have to rename the main PDE file also:

```
CC_Snake/CC_Snake.pde --> CC_Snake_Processing/CC_Snake_Processing.pde
```

otherwise the Processing IDE will fail to open the sketch properly.

nilsweb:
I tried to make the `repository` field behave more "naturally".
I didn't know about the Processing naming convention, but knowing it now, wouldn't it be better to create separate folders inside `CodingChallenges` for each system? This means that all Processing versions go under `CodingChallenges/Processing` and p5.js versions go under `CodingChallenges/p5.js` (similar to whats currently going on inside the `Tutorials` folder). Otherwise every time a processing version is added, the folder and main file would have to be renamed accordingly. Also, I think it would be easier for future contributors now knowing about the suffixes `_Processing` and `_p5.js`. Folders would be more understandable.

shiffman:
@nilsweb I agree, this sounds like a better plan. Would it be:

```
CodingChallenges/CodingChallenge_03_Snake/Processing/...
CodingChallenges/CodingChallenge_03_Snake/p5.js/...
```

or

```
CodingChallenges/Processing/CodingChallenge_03_Snake/...
CodingChallenges/p5.js/CodingChallenge_03_Snake/...
```

One the one hand I like the first best b/c most people are looking for a specific challenge? But this would actually then require another subfolder for Processing, i.e.

```
CodingChallenges/CodingChallenge_03_Snake/Processing/Snake/Snake.pde
CodingChallenges/CodingChallenge_03_Snake/p5.js/index.html
CodingChallenges/CodingChallenge_03_Snake/p5.js/sketch.js
```

nilsweb:
@shiffman Right, haven't thought about the second subfolder... But at some point, people should use the website to download the code so I think it doesn't matter if we use option two?

shiffman:
@nilsweb yes but for anyone working on adding and testing code to the site not having the folder name match the `pde` file will be a major pain. So that folder is necessary.

nilsweb:
@shiffman I understand. That's why I'd go with this folder structure:

```
CodingChallenges/Processing/CodingChallenge_03_Snake/
    - CodingChallenge_03_Snake.pde
    - ...
CodingChallenges/Processing/CodingChallenge_03_Snake/
    - index.html
    - ...
```

Therefore, we don't have any "ugly" folders inside the actual CC folders (which I think would be more counter intuitive than having to decide on the code variation first) and Processing is happy having a main file matching the folder name.

Merijn-DH:
I don't know if my opinion is of any value here, but having a folder structure of `CodingChallenges/Processing/` and 
`CodingChallenges/p5.js/` does not sound logical to me. 
People looking for a specific coding challenge will get confused over which folder it is in and it won't be clear they are available in both versions (of course this only happens when browsing the github and not the website, but I think the github should stay equally usable as the website).

Myself, I'm in favor of either using this stucture: 
```
CodingChallenges/CodingChallenge_03_Snake/Processing/Snake/Snake.pde
CodingChallenges/CodingChallenge_03_Snake/p5.js/index.html
CodingChallenges/CodingChallenge_03_Snake/p5.js/sketch.js
```

even though it requires an extra subfolder, or using the stucture that is used right now..

shiffman:
I think I agree with @Merijn-DH here. @meiamsome is usually the voice of the reason, what say you?

GreyWizard1955:
Guys (especially Daniel, Merijn and Nils),

We're getting to the point that we're already at 3 digits for LSs and close to 3 digits on CCs... Shouldn't we rename all of them to 3 digits before we get there??? JMHO…

Sent from Mail for Windows 10

nilsweb:
Sorry @shiffman and @Merijn-DH but for some reason GitHub didn't notify me about the ongoing discussion...

@Merijn-DH You're right, that would be even more confusing and the GitHub repository should definitely stay useable (and logical/easy to navigate) as well.
I gave it a little more thought and came to the conclusion that the suggested structure would be best. As a reminder:

> ```
> CodingChallenges/CodingChallenge_03_Snake/Processing/Snake/Snake.pde
> CodingChallenges/CodingChallenge_03_Snake/p5.js/index.html
> CodingChallenges/CodingChallenge_03_Snake/p5.js/sketch.js
> ```

Therefore, the repository stays usable and the code people are looking for should be easy to find. Moreover, it's better for contributors when porting a coding challenge as well. If we would decide on the `_Processing` convention, people would also have to think about renaming their main Processing file so everything stays functional at that front. With the convention described by @Merijn-DH, this possible point of failure goes away.

BTW, the "ugly-folder-problem" I mentioned before is irrelevant since GitHub automatically "jumps" into a folder if there is only one subfolder. Haven't thought about that either...

If everyone agrees, I think it would be best to close/reject this PR and I'll open a new one referencing this conversation with the correct changes as it is easier for me to revert the commits and start from scratch anyway.

@blinkityblank Are you done with migrating all coding challenges? I'll wait with my second PR until you're done as we may get *a lot* of merge conflicts if those files change at the same time.

@GreyWizard1955 Could you please create a separate issue about this since it is a little off-topic and it may get lost otherwise?

shiffman:
@nilsweb all sounds good to me!

GreyWizard1955:
@nilsweb I posted in the Slack Channel as well… I just want it looked at… TYVM!!!

Sent from Mail for Windows 10

PR #386, opened 2018-01-16
Swift implementation of the first Coding Challenge

bobvoorneveld:
I thought I would be a nice idea to create Swift versions of each Coding Challenge. So I started with the first one and made it available on Github. Hope you enjoy it!

### DIFF
--- a/_CodingChallenges/01-starfield.md
+++ b/_CodingChallenges/01-starfield.md
@@ -6,10 +6,12 @@ video_id: 17WoOqgXsRM
 repository: /CC_01_StarField
 live_example: /CC_01_StarField_p5.js
 
-  
+- title: "Your version's title (choose a nice one!)"
+  author:
+    name: "Bob Voorneveld"
+    url: "https://www.bobvoorneveld.nl"
+  source: "https://github.com/bobvoorneveld/Coding-Challenges/tree/master/CC001-Starfield"
 
-
-  
 ---
 
 This is the first in a new series of videos.  In this video I attempt to program a "star field" or "warp speed" visualization in Processing.  


_CodingChallenges/01-starfield.md
```diff
@@ -6,10 +6,12 @@ video_id: 17WoOqgXsRM
 repository: /CC_01_StarField
 live_example: /CC_01_StarField_p5.js
 
-  
+- title: "Your version's title (choose a nice one!)"
```
Merijn-DH: I don't think you meant to leave this here?

Merijn-DH:
I don't think you meant to leave `"Your version's title (choose a nice one!)"` in there?

bobvoorneveld:
oops, added the title

Merijn-DH:
Sorry I didn't notice the first time, but as per the wiki:
> Only `title`, `author.name` and `url` are required!

So probably you want to use `url` instead of `source`, for the link to GitHub, which is confusing because the wiki says `url` is for a running example. I'll check with @nilsweb (the creator of the system)

nilsweb:
@Merijn-DH That’s because every link (doesn’t matter in which “section” you use it) in theory supports all options. `url` is needed because otherwise the link would point to nowhere. I’ll definitely have to update the documentation so this becomes clearer.

Merijn-DH:
@bobvoorneveld so, if you change it to `url` all should be good to merge!

bobvoorneveld:
finally got to it, bit of a busy week. Gonna add another pull request for challenge 2 and 3, got them done also.

Merijn-DH:
great!

PR #388, opened 2018-01-16
Port of CC_90_dithering to p5.js

Zalastax:
Hope it's OK that I put my own style on this, since it's not possible to copy the code line by line anyway.


CodingChallenges/CC_90_dithering_p5.js/sketch.js
```diff
@@ -0,0 +1,93 @@
...
+      const clr = getColorAtindex(img, x, y);
+      const oldR = red(clr);
+      const oldG = green(clr);
+      const oldB = blue(clr);
+      const newR = closestStep(255, steps, oldR);
+      const newG = closestStep(255, steps, oldG);
+      const newB = closestStep(255, steps, oldB);
+
+      const newClr = color(newR, newG, newB);
+      setColorAtIndex(img, x, y, newClr);
```
shiffman: I wonder if it would make more sense to just pass in the rgb values instead of making a color object, i.e.

```javascript
setColorAtIndex(img, x, y, newR, newG, newB);
```

CodingChallenges/CC_90_dithering_p5.js/sketch.js
```diff
@@ -0,0 +1,93 @@
...
+  // Apply gray filter to the whole canvas
+  filter(GRAY);
+}
+
+function imageIndex(img, x, y) {
+  return 4 * (x + y * img.width);
+}
+
+function getColorAtindex(img, x, y) {
+  const idx = imageIndex(img, x, y);
```
shiffman: I tend to not use `const` in my examples to be more beginner friendly, but I don't feel super strongly about this. Does it matter much?

CodingChallenges/CC_90_dithering_p5.js/sketch.js
```diff
@@ -0,0 +1,93 @@
...
+  const pix = img.pixels;
+  const red = pix[idx];
+  const green = pix[idx + 1];
+  const blue = pix[idx + 2];
+  const alpha = pix[idx + 3];
+  return color(red, green, blue, alpha);
```
shiffman: Ah, maybe it's good to use the color objects since you are returning something here. . . . this is a good way to organize the code for sure, but would it be simpler for a user to understand if this code were inside the for loop itself. I think my comments should actually be ignored and we should merge this thank you so so much for the help!

CodingChallenges/CC_90_dithering_p5.js/sketch.js
Zalastax: I just use const by default, no problem changing them all to let instead.

Zalastax:
If I understand the review by @shiffman correctly, the only change that should be made is to change const for let. Is that correct?

shiffman:
@Zalastax yes, apologies for stream of consciousness commenting. I tagged @meiamsome to take a look but happy for this to be merged!

shiffman:
I'm going to go ahead and merge this so that it's available, we can always make further changes if necessary. Thank you!

brkus:
Hi I've been trying to make this myself in p5.js but am having trouble with distributing and adding error, when I look at this code I noticed lines 61-63(clr.setRed, clr.setGreen, clr.setBlue) isn't defined so it will come up with an error how would you change this to make it work? Thanks.

Zalastax:
The set-functions were added to p5.js very recently, so make sure you have the latest version of p5.js. Does that answer your question?

brkus:
Ahh that makes sense then. Yes it answers my question thank you.

Issue #370, opened 2018-01-11
Menu links don’t hide on mobile


nilsweb:
Just so I don’t forget (or if somebody finds the issue straight away):

I’ve just noticed that the links in the top navigation bar don’t hide on mobile devices. They are just hovering over the page content even though they should only be clickable when the menu is extended.

I definitely thought that this worked before but I don’t think anything changed in the code since then... Maybe I haven’t noticed?

shiffman:
Could it be when I removed the header image? https://github.com/CodingTrain/Rainbow-Code/commit/0fe45d83dab9071fb0646f57f92752ad4915be94

nilsweb:
Hmm... You just commented out the image tag... That shouldn’t break it but I haven’t had the time to look into this more deeply. Will definitely happen over the weekend.

Link2Twenty:
Works fine for me (google pixel), the sass looks fine too.

Maik1999:
I added a visibility:hidden to the links for mobile and if the checkbox is checked I set them to visible.
But I am not able to get it in sass and test it right now.

PR #392, opened 2018-01-18
🐛  Menu links don't hide on mobile

nilsweb:
Closes issue #370

This should now be fixed. I somehow forgot to set `overflow: hidden` for the `nav` container.
Therefore, the links were "peaking out" since their `height` property was set to `auto` (used for animating the menu in).

Big thanks to @Maik1999 for pointing me in the right direction!


shiffman:
Awesome work!

PR #393, opened 2018-01-19
Swift implementation of the Menger Sponge (CC #02)

bobvoorneveld:


Merijn-DH:
@bobvoorneveld please take a look at commit 83c37f4. I'm sorry we didn't notice earlier, but we're all kind of new to the system.

bobvoorneveld:
updated (finally) with the contributions label

Merijn-DH:
Great!

PR #394, opened 2018-01-19
Swift implementation of The Snake Game (CC #03)

bobvoorneveld:


Merijn-DH:
See my comment in pr #393

bobvoorneveld:
updated (finally) with the contributions label

Issue #354, opened 2018-01-06
Landing Page Design


shiffman:
The landing page should include some elements from current http://thecodingtrain.com and more:

* patreon link
* subscribe link
* coding train merchandise store (shirts / mugs)
* coding train amazon shop

Also, I don't know that the full logo needs to take up so much space might be nice for people to see links to the videos, code, etc. without having to scroll down?

shiffman:
Also add to landing page:

* Next live stream
* Latest coding challenge

Merijn-DH:
Maybe something extra that pops up when you're live? Could be a youtube player or just a 'live' button thats only there when you are streaming?

shiffman:
We can borrow from the current http://thecodingtrain.com/ in terms of the "latest video" design.

Link2Twenty:
There are 3 ways to go about this.

1. Make a `Rainbow-Code/_jekyll/includes/2-base/landing.html` file and include it on the first page.
2. Edit `Rainbow-Code/FrontPage.md` to include.
3. A mixture of 1 and 2.

I think approach 3 is best

Landing | FrontPage
--- | ---
patreon link| merch store
subscribe link | amazon shop
Next live stream | Latest coding challenge*

_*The coding challenges section already has this._

Something like this would be a good start. Though below a certain screen sizes things would have to be done differently, the current breakpoints can be used to make a different layout.

![landing](https://user-images.githubusercontent.com/3534427/34909381-a0118edc-f897-11e7-8e05-ea9626f830ee.png)

nilsweb:
@shiffman What do you think about this layout/mockup?

![6becbe17-d2c4-45d8-894a-d39cac25e065](https://user-images.githubusercontent.com/12577464/35186154-ade86b70-fe0f-11e7-9d80-2b7ad82542bb.png)

shiffman:
This is looking great @nilsweb thank you! Some comments:

* maybe the logo stuff can be incorporated into the navbar? (to conserve space?) maybe not necessary.
* Instead of having a bio or anything on this page I'd rather just add an "about".
* We should add a link to the [amazon shop](https://www.amazon.com/shop/thecodingtrain)

![home](https://user-images.githubusercontent.com/191758/35195258-d0241ed0-fe8e-11e7-8412-aa4ad0b09f12.png)

Would love for any visual designers to weigh in!  (If it's helpful I have all the logo stuff as vector files I'll share via slack.)

nilsweb:
[...]

PR #396, opened 2018-01-19
Issue 355 - Minor front page text alignment, visual tweaks and spelling corrections

davenewt:
- paragraph line spacing increased a touch to aid readability
- links made $ct-purple instead of $text-color to make them more visually distinct on the page
- default front page text-alignment changed to 'left' as per Dan's suggestion in one of the GitHub issues for the top text
- specific pieces of text aligned 'center'
(could have done it the other way around, but left-alignment seems more logical as a default?)
- added link to YouTube CT channel in top text

### DIFF
--- a/FrontPage.md
+++ b/FrontPage.md
@@ -2,35 +2,39 @@
 layout: base
 permalink: /
 style: >
-  text-align: center
+  text-align: left
 ---
 
 <br>
 
 ***
 
-**Welcome to TheCodingTrain's code repository!**
+# Welcome to The Coding Train's Code Repository!
+{: style="text-align: center;"}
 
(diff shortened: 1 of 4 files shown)


nilsweb:
Thanks @davenewt! I closed your other PR and moved title/description to this one since you accidentally created two.

There are definitely some things that I want to bring into the final design. However, I'd wait with merging since I proposed a complete redesign of the landing page on #354. In the case that this takes longer than expected, we could also merge it in as a temporary solution?

One more thing: Setting a different text color for links is a good idea but I don't think it should be `$ct-purple` since this looks very similar to the default "already-visited-link" color used by most browsers. Maybe another one would fit better?

PR #397, opened 2018-01-19
Fix scope of attempts into Coding Challenge 50

nucliweb:
Change the scope of `var attempts` to global into **Processing** and **P5.js** Coding Challenge 50's files.


shiffman:
I believe `attempts` was intended to be local to the `draw()` loop to give the sketch 1,000 chances *each* time through draw to find a spot for a new circle. Is there something I'm missing that requires it to be global?

nucliweb:
In the block while `attempts ++` it will never reach 1000, since when `count` reaches 9 it leaves the while block, and in the next iteration of the `draw()` function the variable `attempts` is again 0.

So, `if (attempts > 1000)` *(with `noLoop()`)* it will never true.

shiffman:
Yes, I understand. This is in fact the case as I don't want it to be true! It's just protection against an infinite loop in case it tries for 1,000 times and doesn't find a free spot. Thank you again for the contribution!

nucliweb:
Ok, I get it, thanks

Issue #400, opened 2018-01-21
Deprecate Coding Challenge README's


shiffman:
Following up from #389, we can now remove the community contributions from the Coding Challenge README's and add a note to each one how to contribute with the new system.

shiffman:
Noting that as each one is done, the community contributions need to be copied over to the new markdown file associated with each challenge!

Merijn-DH:
Is it ok if I assign this to myself? I'm done up to CC 68 at the moment.

shiffman:
Yes, of course, thank you @Merijn-DH!

Merijn-DH:
I'm so sorry I have been inactive for the past few weeks, I'll get to this very shortly!

Fraasi:
This transition is long done and issue could be closed.
@shiffman Could you confirm?

shiffman:
indeed!

PR #402, opened 2018-01-21
Try-out for converting community contributions to the new website

Merijn-DH:
Referring to issue #400,

Thought it'd be good to try one out before I start bulk-moving them and find out I did something wrong. Let me know if this looks good or if anything should be different.


Merijn-DH:
Great!

PR #431, opened 2018-02-11
Livestream #116.2

N4G1:
Livestream #116.2 updated and fully linked.


_Streams/116.2_-_Snakes_&_Ladders_Continued_+_2048.md
```diff
@@ -0,0 +1,30 @@
...
+date: 2018-01-26
+video_id: 1v_E18xU3ok
+
+topics:
+  - title: "Continue Coding Challenge: Snakes & Ladders"
+    time: "9:30"
+    url: "/CodingChallenges/91.3-snakes-and-ladders"
+  - title: "Coding Challenge: 2048 in Processing"
+    time: "56:33"
+    url: "/CodingChallenges/"
```
nilsweb: I think there's something missing here. 😉

N4G1: I left it "open" because the video is not uploaded yet, so i wrote it halfway so we can add the video sligtly faster when it comes up. I wanted actually to ask You about it, what is the procedure when the video is not up yet, do we just leave the "" empty or?

nilsweb: Just remove `url` altogether when the video hasn't been uploaded. Then Jekyll won't make a link and just use static text instead so people aren't confused why the link isn't going where it should.

N4G1: Done, i changed it in my previous pull-requests too ;)

PR #434, opened 2018-02-11
Livestream #117.2

N4G1:


nilsweb:
A big thank you for migrating all the streams! This helps us a lot! 👍 
Just a short note: I think that there are multiple parts of stream 117. That should also be reflected in the `video_number` and title (with something like `...-part-1.md`, `...-part-2.md`).

N4G1:
@nilsweb You are right, there were two livestreams 117.1 and 117.2, but 117.1 lasted for 15 min and Dan had to go, and when he came back he did 117.2 and repeated all he said in 117.1 thus i thought that i will just migrate them as one livestream.

nilsweb:
@N4G1 That's true but I think it would be confusing for others if the video numbers suddenly don't match anymore. Waiting for @shiffman's opinion on this.

shiffman:
Yeah, I'm not sure what's best here. . but since, at the moment, we're mirroring every separate video on YouTube with a page here then I think we should have a `-1` and `-2`?

nilsweb:
Thank you very much for your help, @N4G1!

PR #435, opened 2018-02-11
Livestream #118 update

N4G1:
Updated filename from _ to - in order to keep filenames consistent, and removed empty ulr's.


nilsweb:
@shiffman I think that the coding challenge on XOR was remade in stream 119. Should this link to the new video or to nothing?

N4G1:
It is linked to the new XOR video.

nilsweb:
@N4G1 I know, but it might be confusing for some people that if you click on the link here that it will take you to a video that's different from what you've seen in the stream.

N4G1:
@nilsweb I agree but at the same time if someone would like to follow along it would be better for him to use the newer version since it is better done (there is a reason why Dan is doing the videos all over again). Thats my humble opinion, but You can decide, and I will follow whatever You chooses in my future run.

nilsweb:
I get your point @N4G1, makes sense. Just waiting for @shiffman's opinion on the topic.

shiffman:
I don't think it's a big deal either way. . let's have it link to the new XOR video as most likely if someone is looking for that content they'll be happy to have found their way to the right video.

PR #449, opened 2018-02-13
fixed mispositioned background in CC93

mikakruschel:


mikakruschel:
Before the fix, it looked like this:
image

PR #457, opened 2018-02-16
Update sketch.js

dancojocar:
Add number only if there are spaces left


MatthewBraun:
Great catch :D I've been working on the sketch for a couple hours now and didn't notice this.

PR #459, opened 2018-02-17
Update sketch.js

Un-knownCoder:
I added this function to check if the player has reached the number `2048`


shiffman:
Nice!

shiffman:
(Just to note that typically I don't add new features to the code examples, but this is such a small feature that doesn't really affect any of the code from the tutorial so great to have.)

PR #462, opened 2018-02-17
Create 94-2048

jelitter:
My second attempt ! Sorry I'm pretty new to collaboration in Github =)

### DIFF
--- a/_CodingChallenges/94-2048
+++ b/_CodingChallenges/94-2048
@@ -1,0 +1,14 @@
+---
+title: "2048"
+video_number: 94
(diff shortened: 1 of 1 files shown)


_CodingChallenges/94-2048
```diff
@@ -0,0 +1,14 @@
+---
+title: "2048"
+video_number: 94
+date: 2018-02-16
+video_id:
```
nilsweb: This will cause problems when Jekyll tries to build the site. Please put the video's ID here (when it gets released).

PR #463, opened 2018-02-17
Create 94-2048.md

jelitter:
Left empty `video_id` since the video isn't uploaded yet


jelitter:
image

nilsweb:
@jelitter If you create a file for a video that hasn't been published yet, just omit the `video_id` property completely. That way, it can be merged but it won't show up on the page. If you leave it like you have it now, Jekyll expects some form of value for `video_id` and will give errors if there isn't one.

PR #466, opened 2018-02-18
Minimal code refactoring

Un-knownCoder:
Hey, I tried to make some code refactoring... hope you need this.

> P.S. I'm not english, so there could be some language errors in the <b>comments</b> of the file


Un-knownCoder:
Oh, before I forget it, there is a small bug, I think, on the `blankGrid()` function, because if you try to create the array with `let #name = new Array(4) ...` the program does not work as it should
> (this I noticed on a WINDOWS computer)

Issue #417, opened 2018-02-04
Unit Testing and Continuous Integration


shiffman:
Now that I have done some tutorials about unit testing and continuous integration, perhaps it would make sense to write some tests to make sure the jekyll build isn't broken and/or any markdown pages contain all the necessarily fields?

I know that @haideralipunjabi created #326 which may be relevant?

nilsweb:
This sounds like a really good idea!
Unfortunately, I haven't worked with unit testing and CI yet (and also wasn't able to attend your last stream 😞) but that way I'll learn something as well.

haideralipunjabi:
I haven't been in touch with your channel (coding in general) for the past few weeks. I will make sure to see the latest live stream. 
#326 shouldn't mess with Jekyll as it only modifies the README.md.

nilsweb:
I've collected some ideas for test cases that should help verifying pull requests. We should check for:
- general Jekyll build errors
- valid URLs (to external sources, other videos, etc.)
- correct naming conventions (especially for the filename since it is used to build the URLs)
- missing properties (e. g. every contribution has to have an author and a URL; all videos have to have a date, title and number)
- use of `video_id` over a YouTube URL to enable automatic matching at a later point
- use of an internal URL (like `/CodingChallenges/...`) over `video_id` when the referenced video is already available on the website
- duplicated videos (could happen accidentally)

As soon as I notice something else that could be tested automatically, I'll update the list.

arisanguinetti:
@nilsweb do you think you can convert your list to be checkboxes that can be thicked?

shiffman:
Can this be closed with #471 and #501? New Issues can be opened for specific testing bugs and features?

meiamsome:
I think I'd like to keep this open until more of the issues in @nilsweb's comment are resolved.

haideralipunjabi:
I have zero experience with node, but hopefully I can contribute my ideas.
The valid url issue could be resolved by checking the status code each url returns, directly or using a service (if there is any?).
The naming convention could be achieved using RegEx.

nilsweb:
I don't know if checking the status code for each URL is practical since there are a LOT of URLs to test. The tests would take a ridiculous amount of time to complete. Maybe we could use a RegEx here as well just to make sure that it *might* be a valid URL?

[Nils - Contact Using Hop](http://GetHop.com/?_hmid=1529180341)

[...]

PR #471, opened 2018-02-20
Add test platform and a simple test to check links in .md files

arisanguinetti:
I've added the jest test platform and also a simple test to check links in all .md files.

I'm sure this could be improved since I'm a beginner in node and jest ;)

Also, keep in mind this test could take some time to run since it will check all links in all .md files.

I think this was on @nilsweb list on #417...

### DIFF
--- a/__tests__/check-urls-in-md-files.js
+++ b/__tests__/check-urls-in-md-files.js
@@ -1,0 +1,74 @@
+const fs = require('fs');
+const path = require('path');
+const markdownLinkCheck = require('markdown-link-check');
+
+let walk = function(dir) {
+  let results = []
+  let list = fs.readdirSync(dir)
+  list.forEach((file) => {
+    file = dir + '/' + file
+    if (dir.replace(/\.\//g, '') === '.git') return;
+    if (dir.replace(/\.\//g, '') === 'node_modules') return;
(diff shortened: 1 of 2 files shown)


arisanguinetti:
Forgot to mention that the test is currently not passing since there are many broken links and I was not sure what to do about it (i.e. Should we remove controbutions with a broken link or just remove the reference?)

nilsweb:
This looks very interesting! Thank you! I'll look more into it over the next couple of days but it seems like a good starting point.

arisanguinetti:
@nilsweb thank you! I've taken this approach due to @shiffman's testing videos but, since there's ruby in the repo for Jekyll, tests might be performed using ruby instead of Node.

If you decide to go with Ruby, then we might need to see this gem:
[https://github.com/gjtorikian/html-proofer](https://github.com/gjtorikian/html-proofer)

nilsweb:
Personally, I'd go with Jest anyway since I don't really like Ruby at all and we're just checking content, conventions, etc. and not actual code.

shiffman:
@arisanguinetti thank you so much for this contribution! I have merged #493 which included the full circleci integration in lieu of this one, but I would love for you to contribute there. Thank you again!

PR #477, opened 2018-02-21
Adding my own version of this Coding Challenge

kim-marcel:


meiamsome:
Ok, I've merged this but please note that we have changed contributions to a newer format so they are standardised and on the website. Feel free to submit another PR adding them there, otherwise it will be caught when these contributions are cleaned up!

PR #464, opened 2018-02-18
Livestream #121

itsTurner:
This is what I could get done in an attempt to get live stream #121 on the website:

* Update Markdown file name
* Update live stream title
* Omit time from date as per contribution guide
* Added video_id
* Most other things mentioned in contribution guide

PR #479, opened 2018-02-22
Adding livestream 120, 121 and ES6 series .5, .6, .7

ZwergB:
I've seen #464 and his version of livestream 121 is pretty much the same as mine, but new videos made from it are already linked in my file.


nilsweb:
Thank you very much @ZwergB! This helps a lot!

PR #486, opened 2018-02-28
Sentence correction

akanksha1612:
Sir i find that writing in this way will be more easy for understanding.


shiffman:
Thank you!

PR #488, opened 2018-03-01
Please add my custom version of 2048

rajatkantinandi:
## 2048-Rapid-Mode

Classic 2048 game in rapid mode with chain reaction type addition. 
Created in javascipt using jquery framework & juery touchSwipe add-on for swipe events in mobile browser.

Supports both swipe & arrow keys.


shiffman:
nice work!

PR #489, opened 2018-03-01
Update 93-double-pendulum.md

grilly86:
please consider adding my example, which I could not make without Dan's great explanation!


shiffman:
Awesome work!!

PR #492, opened 2018-03-03
Add contribution to #51

donwilson:
Added in a little ES6 + restart with new random map a moment after the algorithm completes


shiffman:
Nice!

PR #494, opened 2018-03-05
Fixed incorrect video id

ZwergB:
It was missing a lil dash


nilsweb:
Thank you for noticing!

Issue #496, opened 2018-03-05
Adding a linter


MeaningOf42:
As @shiffman mentioned in #493, it would be a good idea to add a linter to this site to ensure only clean code is submitted. I would be willing to give this a go, but I have some questions about his comment:

>I prefer spaces around => I guess I should add some sort of linter to this repo? Is AirBnB style guide what I should adopt and is there a linter that goes with it? This is new to me, but I would love to make a video about it so advice appreciated! @meiamsome

+ As Daniel asked, would the AirBnB style guide be what we should adopt? @meiamsome 

+ ESLint seems like a good linter to use with JavaScript and it a fair number of other people have documented using it to enforce the airbnb style guide therefore it seems like a good idea to use. Anyone have any ideas regarding the choice of linter we use?

+ I think we want to do the linting as part of circleci to prevent unclean code from being committed, am I right in this matter.

+ In "I guess I should add some sort of linter to this repo?" by this repo do you mean the whole website or just unit testing?
  + If you mean the whole site then you might have to change tutorial code to comply with the style guide, which would mean deviating from the code shown in the videos, would this be ok?

meiamsome:
I linked Dan to ESLint's [website](https://eslint.org/) as my suggestion for a linter as it is my preferred JS linter. The init of ESLint lets you chose from 3 default styles or create your own. The defaults are:
1. [Google](https://google.github.io/styleguide/jsguide.html)
2. [AirBnB](https://github.com/airbnb/javascript)
3. [Standard](https://standardjs.com)

Note that Standard disallows semicolons at the end of lines, which is very contrary to @shiffman's style as I have seen it. I could not work out if Google or AirBnB matches closer. The other option is to just create an ESLint config manually, which is not terribly hard as ESLint defaults are fairly non-controversial. We also have to note that there is a mixture of Node and Browser code in this repository if we want to lint _everyting_, so we need to be careful about what we disallow (Some older style code produces warnings in the standard style).

MeaningOf42:
I dislike the idea of starting a new config from scratch as that would create a new style for people to learn. However if we made sure that all our decisions are compatible with one of the style defaults then, I would be ok creating a custom config file, as we could simply tell people it is a less strict version of the style it is compatible with. (Not sure if I was clear. My problem would be if we adopted some practices from Google and some from AirBnB as this would create a confusing mixture of the two.)

[...]

PR #503, opened 2018-03-07
Starting to add eslint #496

shiffman:
This is not at all ready to be merged. I am learning about how linting works and am starting this branch to sort it out. Some questions:

1. I have a `package.json` now in the root directory. Should this be moved to a `linting` directory or merged with`unit_testing/package.json`?

2. As suggested by @Versatilus I am starting with the p5.js `.eslintrc.js` configuration. This can change over time. This is what I'm seeing, however. How should I deal with these "undefined" elements that are part of the p5.js library?

screen shot 2018-03-07 at 5 28 14 pm

3. How do I properly set-up the linter to run? Should it be part of testing / circleci? Should I only be running it on `js` files or should I set something up for other files too? (`html`, `css`, `pde`, `json`, . . . )

Tagging @meiamsome for feedback.


meiamsome:
Looks like you can add https://www.npmjs.com/package/eslint-config-p5js to packages & then to ESLint config to let it find the globals.

As for how to deal with merging with unit testing, I think we should move the unit testing functionality into the main package.json so that it's all in one place - I'm happy to do this maintenance after we have the linter setup in the root file. To get linting on the CI we'll probably want to make the npm test command run multiple things, possibly with https://www.npmjs.com/package/npm-run-all ? I can deal with this too when merging the packages, if you want?

Versatilus:
The first step is to install the `linter-eslint` package in Atom. Also change the beautifier to use `ESLint Fixer` for JavaScript.

The settings for eslint cascade down through the directory tree, with higher priority going to configuration files closer to where you're working. In the root of whatever repository you're working in, you should add the wanted components to the package.json as developments dependencies.
`npm i —-save-dev eslint eslint-config-prettier eslint-plugin-prettier prettier`

In the `.eslintrc` file you're going to want to remove `node` and `mocha` from the `env` section. They can be added back where relevant. Particularly, the definitions for node interfere with the way I recommend handling eslint not recognizing p5 symbols. ESLint can read settings from comments. Just like I did with my neural network code, you're going to want to add two comment sections to manage that. `/* globals */` lists what you bring in from other scripts and `/* exported */` lists what you define for the outside, like `setup` and `draw`.

I prefer this method over adding `eslint-config-p5js` simply because it provides documentation in the code of what's coming from where. If you go that route instead, feel free to leave `node` in `env`.

Versatilus:
Feel free to ignore that ranting about `node` and `mocha`. I didn't realize you were using the file I posted.

shiffman:
Going to attempt to install the `linter-eslint` package in Atom and change the beautifier to use `ESLint Fixer` now too, thanks @Versatilus!

gruselhaus:
@shiffman Just for clarification: Do we want eslint just for website related js or for "CT-Code" as well? I run eslint with the current p5 config and encountered 10.000 problems. I double checked and restricted to website related code only and found 18 problems. I guess that the  "CT-Code" causes the most problems.

jonnytest1:
how about a precommit hook that checks all the changed files for linter errors so you can only commit if your changed files fit the linter

shiffman:
@gruselhaus great question! I'm not sure! I think my goal here is just to establish some conventions and keep the code as clean / consistent as possible.  I am thinking about this for the  code examples themselves! But I am also open to the website code as well. . but these would likely need different linting as I am pretty loosey-goosey with my code?

jonnytest1:
@shiffman for a lot of rules eslint adds auto formatting (with vscode support) you can tick auto format on save and whenever you save it reformats your document to fit the rules (I use this feature a lot even while developing)

gruselhaus:
Added in #1929

Issue #499, opened 2018-03-06
Seems to be a problem if more than one series in "Courses" or "Tutorials".


blinkityblank:
I started adding the "Learning Processing" series to the "Tutorials" section, but there seems to be a problem with directories within directories.
I put the JavaScript and p5.js tutorials in their own series and started a "Learning Processing" series.
On the Tutorials page, everything looks great:
https://blinkityblank.github.io/Rainbow-Code/Tutorials/

But if you click on the "Watch all sessions" for the "Learning Processing" tutorial, it partially displays the chapters from the JavaScript and p5.js as well.
You can see it on a gh-page from my repo: 
https://blinkityblank.github.io/Rainbow-Code/Tutorials/learning-processing/

![image](https://user-images.githubusercontent.com/6036839/37047905-6134dfa4-213a-11e8-9361-542dad7f429b.png)

Maybe I didn't configure something properly, but I believe I did. The problem basically happens when you have a series folder within a series folder. Is it supposed to be possible to do so and it's a bug or the website is not supposed to be working like that.

Thanks!

nilsweb:
This is somewhat intentional. "Learning Processing" is actually a "series of series" (or "playlist of playlists" if that makes more sense). A tutorial is simply a playlist. Thats's why the website behaves weirdly if you mix those two together.

I would think of "Learning Processing" as a course instead of a tutorial, similar to Programming with Text or The Nature of Code. Try to add it to the `_Courses` folder istead. It should look something like this:

```
_Courses
  L learning-processing
    L 0-introduction
      L 0.0-introduction.md
      L 0.1-who-am-i.md
      L ...
    L ...
  L programming-with-text
    L ...
```

That way, it should work as you intended. Hopefully that explanation makes sense.

blinkityblank:
I agree that "Learning Processing" should be considered a course, since it's pretty focused on one big subject.
But even if I put it in the Courses folder, it seems the be behaving the same way:
https://blinkityblank.github.io/Rainbow-Code/Courses/learning-processing/

And I believe I respected the hierarchy that you posted above?

nilsweb:
That's weird... I honestly don't know what's causing this. Everything you did seems totally fine. Maybe that's a bug? I'll look into it, might be a fault on my side... 😬

blinkityblank:
I think i figured out why! To see if a series was to be added, it only if it was in the right collection("Courses") and the number of slashes(to check if it was at the same level of the hierarchy).
[...]

PR #504, opened 2018-03-08
Added Learning-Processing Course and fixed the "series within series" problem #499

blinkityblank:
Added Learning-Processing Course with Session 0.
Also Fixed the "series within series" problem from issue #499

Instead of looking at the number of "/" to see if it's at the same level in the same collection, I make it look at the whole path.


blinkityblank:
Great, merging!

PR #510, opened 2018-03-11
CC_94_2048: Removed unnecessary direction in transposeGrid

KebapBoy:
I've removed the unnecessary `direction` argument from the `transposeGrid` function
because transposing a matrix is not dependent on a direction and this might be irritating.


shiffman:
Thanks for this. If I ever make a part 5 someone help me remember to mention this mistake that I made!

PR #511, opened 2018-03-11
Multi-part videos should only allow contributions to the first part

nilsweb:
As requested by @shiffman and @meiamsome, contributions on multi-part videos are now only allowed for the first part (video `XX.1`). Each video in a multi-part series now lists the contributions made in the first video instead of its own, unless specified otherwise. To control how contributions should behave, refer tho this set of rules (only applies to multi-part videos):
- If a video doesn't contain the `contributions` property it shows the contributions made to the first video in the series (video `XX.1`).
- If a video already contains the `contributions` property it shows its own contributions.
- If a video should allow direct contributions, set `contributions: false`

### DIFF
--- a/_jekyll/layouts/video.html
+++ b/_jekyll/layouts/video.html
@@ -67,14 +67,27 @@ layout: base
     {% comment %} List of community contributions (only visible if can_contribute is true). {% endcomment %}
     {% if page.can_contribute %}
       <div class="contributions">
+        {% comment %} Converts page.video_number to a string. {% endcomment %}
+        {% capture video_number %}{{ page.video_number }}{% endcapture %}
+
+        {% comment %} Set video which contains the contributions. {% endcomment %}
+        {% comment %} Multi-part videos allow contributions only on the first part. {% endcomment %} 
+        {% if video_number contains '.' and page.contributions == null %}
+          {% assign video_number_base = video_number | split: '.' | first %}
+          {% assign firstPartVideoNumber = video_number_base | append: '.1' %}
+          {% assign contributionsVideo = site[page.collection] | where: 'video_number', firstPartVideoNumber | first %}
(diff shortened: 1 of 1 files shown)


_jekyll/layouts/video.html
```diff
@@ -67,14 +67,27 @@
     {% comment %} List of community contributions (only visible if can_contribute is true). {% endcomment %}
     {% if page.can_contribute %}
       <div class="contributions">
+        {% comment %} Converts page.video_number to a string. {% endcomment %}
+        {% capture video_number %}{{ page.video_number }}{% endcapture %}
+
+        {% comment %} Set video which contains the contributions. {% endcomment %}
+        {% comment %} Multi-part videos allow contributions only on the first part. {% endcomment %} 
+        {% if video_number contains '.' and page.contributions == null %}
```
meiamsome: What's the purpose of the `page.contributions == null` check here? If we want to enforce this can we not just ignore the current contributions (And I can lint them as invalid and move invalid ones to their master file)

nilsweb: I thought that it would be nice if we can still have contributions on 'sub-parts' when needed. For example, a challenge might have five parts where part one to four is the challenge itself and part five is a variation of the original challenge made by @shiffman.

shiffman: Is there an example of this being needed for a particular multi-part challenge at the moment? Might be easier to maintain if we keep it simple and add this feature if it turns out we need it later? I'm ok either way and defer to @nilsweb and @meiamsome.

meiamsome: I think it makes sense that the Jekyll side be programmed for this case either way. We can enforce this in tests instead.

shiffman:
woohoo!

PR #514, opened 2018-03-12
Fixed broken links to coding challange 90.

MeaningOf42:
I was so exited to submit my contribution, I forgot to check the links work. I just fixed them.


shiffman:
Maybe we should add a test to check for this! 😜

PR #515, opened 2018-03-12
Updated for loop to ES6 syntax

spiray:


meiamsome:
Hello,

This repository does normally accept code changes so that the code stays as close to the original videos as possible. We are happy to accept fully fleshed out versions which are code ports (including to ES6) as community contributions, though!

spiray:
@meiamsome understood. Thanks

PR #516, opened 2018-03-13
CC 58 fix coordinate system

micuat:
I fixed a coordinate issue in earthquake viz 3D. I think the issue came from the fact that in opengl coordinate system, y and z are flipped from typical notations of sphere-Cartesian conversion.

I noticed this immediately because I'm from a country of earthquakes :)


meiamsome:
This repository does not usually accept code changes so it matches the videos closer. That said, this is an issue that is noted in the challenge's video and I can see it being very beneficial to be added to the code especially as it appears (At least to me) to be correct! I'll defer to @shiffman for the final say, though!

shiffman:
Yes, I would love to merge this! It's an important fix. Would you mind adding a code comment that explains the fix / difference between the video for context?

micuat:
ok! I added comments and also kept the original code in the comments

PR #532, opened 2018-03-23
sketch.js: no need for played variable

AceVerso:
There is no need for played variable as if any other key was pressed, we can simple return


meiamsome:
I think this one is more stylistic than bug fix, either code is fine so I think it's best to stick with the one from the video.

PR #541, opened 2018-03-26
Creat README.md

Lugsole:


nilsweb:
The `README.md` files are deprecated as discussed in #400 so there is no need to add it. Thanks anyways!

PR #544, opened 2018-03-26
 Coding Challenge 98: Add quadtree.md files and contribution

romibi:
**Added the following files based on info from Youtube videos:**
_CodingChallenges/98.1-quadtree.md
_CodingChallenges/98.2-quadtree.md
(The date in part 2 might need modification once the youtube video is listed.)

**Added my contribution with Mouse-Wheel control to range size**
For live example see https://alpha.editor.p5js.org/full/HkDKgnIcz
For Source see: https://github.com/romibi/p5.js-Playground/tree/master/CC_98_QuadTree

**Open Questions:**
- Are readme.md files in the CodingChallenges/ subfolders with contributions now unwanted or still welcome?
- Is that correct that contributions in these files always have to be in part 1?

### DIFF
--- a/_CodingChallenges/98.1-quadtree.md
+++ b/_CodingChallenges/98.1-quadtree.md
@@ -1,0 +1,23 @@
+---
+title: "Quadtree - Part 1"
+video_number: 98.1
+date: 2018-03-26
+video_id: OJxEcs0w_kE
+repository: CC_98_QuadTree
+
+links:
+- title: "Quadtree on Wikipedia"
+  url: "https://en.wikipedia.org/wiki/Quadtree"
+
+videos:
+- title: "Quadtrees Live Stream Archive"
+  video_id: "MxnqJGwu2cc"
+- title: "Binary Tree Coding Challenge"
+  video_id: "ZNH0MuQ51m4"
+
+parts:
+- title: "Quadtree - Part 2"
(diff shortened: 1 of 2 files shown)


romibi:
of course feel free to only use the second commit and create the quadtree.md files yourself

blinkityblank:
@romibi Thanks for those! Couple of notes that you can change before we merge:

- The links in the "Parts" sections have a mistake.
image
It should be `98.1`, not `97.1`

Also, when the challenge is made with JavaScript, you can remove the line `live_example: false` altogether. The repo is used as the live_example.

You can make those changes and I believe we're good to merge!

romibi:
oh right …
fixed

what about my open questions from the PR's description?

blinkityblank:
Thanks for the corrections.

For the multi-part: Yes, all the contributions only go in Part 1.
On whether or not we still accept README.md files in the Coding Challenges directories, I know that they are not necessary anymore, but I'll let @nilsweb give the official answer on that.

nilsweb:
In my opinion, the readme files should only be used to add notes to the code itself and not for adding contributions, but for an official answer we have to ask @shiffman.

shiffman:
Yes, the contributions should now live in the markdown web pages! Thanks @romibi and @nilsweb!

PR #98, opened 2017-03-02
circles based on pictures

wisamtb1:
I Changed the stroke on each circle, to become a variable - which i then decrement its value to get different color circle per instantiation.                        
change the image background that the circle is packing                        
I Changed the stroke to have a random green hue as well

PR #548, opened 2018-03-27
CC #98 Bug fixes: Quad Capacity & Mouse out of bounds

VyraWasHere:
Fixed two bugs.

Bug: A quad could contain more than 4 points when the parent's points overlaps with it 
Fix: The parent now moves all of its points to the appropriate quad when subdividing.

Bug: Moving the mouse cursor out of range for the root quadtree's boundary would throw a TypeError when qtree.query() returns null.
Fix: Only query the quadtree when the mouse is within the root's boundary.


nilsweb:
It seems like Git messed something up here since it shows over 700 file changes. Could you please resubmit a fix if the bug is still there? Thank you!

PR #550, opened 2018-03-28
Book of Pi with polygons

VegaVen2ra:
Based on Coding Challenge 97.1 I implemented the idea to represent the digits of pi with polygons where the number of edges equals the corresponding digit. [Here](https://github.com/VegaVen2ra/CC_97_Book_of_Pi_Polygons/blob/master/bookofpi-1million-polygons.pdf) I uploaded the created pdf.


shiffman:
love it!

PR #555, opened 2018-03-29
Permanent upcoming stream page

shiffman:
Trying to have a page I can always update and edit to schedule next live stream.


shiffman:
Looks like I can't name the file `999` and bump up the number for each new stream.

I should really run these tests locally. . .heh. .. .

screen shot 2018-03-29 at 5 36 06 pm

nilsweb:
Since the upcoming stream feature isn't used anymore due to issues like this, I'll go ahead and close this now!

PR #556, opened 2018-03-31
Adding safety-check to prevent "points no iterable" error

Carla-de-Beer:
This error occurs when when the mouse position is off-canvas in the CC_98_QuadTree example code.


shiffman:
Thanks for this! I think it makes more sense to correct the bug that is in the `QuadTree` class where found returns `undefined` rather than an empty array `[]`. You can see the corrected code here if you want to pull request the same to this repo:

https://github.com/CodingTrain/QuadTree/blob/master/quadtree.js#L148

nilsweb:
@shiffman Should I resolve the conflicts and merge?

shiffman:
@nilsweb yes, though I think it's probably good policy to add a comment when the code differs from the video challenge?

nilsweb:
I added a short comment as suggested!

PR #559, opened 2018-04-08
Coding challenges 98 and 99

shiffman:
This pull request has the code for challenges 98 (parts 1/2 and 3) and 99. Note that the original QuadTree example has been renamed now that there is a part 3. @blinkityblank perhaps you can update the markdown files to reflect this? While the QuadTree is a multi-part challenge there are two examples that came out of it, one for part 1 and 2 and one for part 3. . is the way I'm uploading this acceptable?


shiffman:
I realize I requested reviews, but I am going to go ahead and merge this b/c I am sharing the code with my students right now and it's easier to find if it's been merged. We can make adjustments after the fact, thank you!

PR #562, opened 2018-04-10
Add a contribution

bibixx:
I have created a version where training data is based on color contrast rather then on a set threshold.


shiffman:
Looks great, thank you!

Issue #574, opened 2018-04-20
CC 93 DoublePendulum p5.js buffer DPI issue


micuat:
I noticed that p5.js version of double pendulum renders incorrectly on my Microsoft Surface (probably on retina macs too). I needed to add `pixelDensity` here:

```
function setup() {
  createCanvas(500, 300);
  a1 = PI / 2;
  a2 = PI / 2;
  cx = width / 2;
  cy = 50;
  buffer = createGraphics(width, height);
  buffer.pixelDensity(1);
  buffer.background(175);
  buffer.translate(cx, cy);
}
```

should I go ahead and make pull request?

shiffman:
ah, yes, thanks for this! Go ahead and pull request! The quality might suffer a bit of a high density display so it might be worth investigating other techniques for drawing the trail at some point... looking forward to your tweet!

Merijn-DH:
Closing :) Further discussion can be done in the pr you made (#575)

fepegar:
Sorry for necrobumping, but I removed the `pixelDensity` line and it looks fine on my retina display. Happy to submit a PR if you like.

micuat:
@fepegar thanks, but it seems like already fixed here https://github.com/CodingTrain/website/pull/737

fepegar:
That PR _added_ the line, I suggested _removing_ it.

PR #575, opened 2018-04-24
CC 93 fix buffer dpi

micuat:
This fixes issue https://github.com/CodingTrain/website/issues/574 - CC 93 (p5.js version)'s buffer size for high DPI screens. I tested it on Surface Pro 4 (175%).

### DIFF
--- a/CodingChallenges/CC_93_DoublePendulum_p5.js/sketch.js
+++ b/CodingChallenges/CC_93_DoublePendulum_p5.js/sketch.js
@@ -28,6 +28,8 @@ function setup() {
   cx = width / 2;
   cy = 50;
   buffer = createGraphics(width, height);
+  // correct buffer size for high DPI displays
+  buffer.pixelDensity(1);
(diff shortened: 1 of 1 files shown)


micuat:
side note: with this fix, the buffer looks blurry as @shiffman pointed out. Better solution is this

```
function setup() {
  createCanvas(500, 300);
  a1 = PI / 2;
  a2 = PI / 2;
  cx = width / 2;
  cy = 50;
  let n = pixelDensity();
  buffer = createGraphics(width*n, height*n);
  buffer.background(175);
  buffer.translate(cx, cy);
}
```

but I found it a bit less intuitive (although it makes crisp lines)

nilsweb:
@shiffman Should I resolve conflicts and merge?

shiffman:
@nilsweb sure, yes, thank you!

gruselhaus:
@nilsweb this was solved in #737

PR #577, opened 2018-04-27
fix title order

v8xm:
According to the url I think the order of Html and Canvas is wrong .


nilsweb:
The `README.md` files are deprecated as discussed in #400 so there is no need to fix the order. Thanks anyways!

Issue #542, opened 2018-03-26
update p5.js version


Spongman:
@shiffman, i just noticed you were having issues with the friendly errors performance in your quadtree live stream. the reason for this is that you're using an old version of p5.js (0.5.16).

the latest version significantly reduces the overhead from **43**% (!!) of total CPU time in 0.5.16:
![image](https://user-images.githubusercontent.com/1088194/37927443-f594202e-30ee-11e8-91af-06801ff82591.png)

to less than **1**% in 0.6.0:
![image](https://user-images.githubusercontent.com/1088194/37927760-ccbfe59c-30ef-11e8-8925-fda7223a30cd.png)

shiffman:
Oh this is so good to know, thank you! Glad to accept pull requests updating versions!

micuat:
actually i noticed CC_02_MengerSponge_p5.js has another issue with p5.js version. it works properly with 0.5.11 which is shipped with the code, but when I updated the library to 0.6.0, nothing was rendered. I learned that I needed to add `normalMaterial()` to render with normal colors.

I wonder if all the codes need compatibility check with the latest p5js or just leave them as they are.

Spongman:
can you try with 0.6.1 (which was just recently released) ?

meiamsome:
Closing as fixed by #581

PR #581, opened 2018-05-02
use CDN for p5.js libs. remove local p5.js files.

Spongman:
re #542

- updates local p5 libs to 0.6.1 CDN versions.
- removes (now) unused p5 files


micuat:
I tested cc 01 and 02 and they work fine. maybe comment like

```
  // as of p5.js 0.6.0, normal material has to be explicitly selected
  normalMaterial();
```

will be helpful.

shiffman:
I'd like to merge this, any comments or questions before I do? @meiamsome @nilsweb

meiamsome:
No issues here! Looking forward to having 10 million less lines...

shiffman:
Yay!

PR #587, opened 2018-05-09
browser version of "2D Water Ripple"

ds604:
used 8 neighbors instead of 4, but otherwise same as Processing version (used to compare speed of canvas vs Processing)


shiffman:
lovely!

PR #601, opened 2018-05-21
Added 103-fire-effect contribution

stellartux:
Added my version of the Fire Effect coding challenge.


shiffman:
Awesome! I have really got to do some shader tutorials at some point!

stellartux:
I would like to see that. I was going to implement the fire effect in p5, but I wasn't too sure how to make the vertex shader or frame buffers.

PR #611, opened 2018-06-03
Added a community contribution to CC64.2

quillaja:
Based on the result Dan shows in coding challenge 64.2 "inverse kinematics".

live example: 
http://www.quillaja.net/snakes/sketch.html

code:
http://www.quillaja.net/snakes/
https://github.com/quillaja/p5-sketches/tree/master/kinematics


shiffman:
Thank you!

PR #612, opened 2018-06-03
Added a community example to circle packing

quillaja:
I implemented an alternative (but not necessarily better) method of circle packing based on a paper by Paul Bourke in which he describes using an infinite function that converges to the area of the space to be filled. I also added a number of bells and whistles in terms of displaying the result. Spacebar will toggle "2d" or "3d" mode--display as circles on a plane or spheres. Arrow keys rotate the result, "R" resets, and "P" toggles between orthographic and perspective perspectives.

Live demo: http://www.quillaja.net/circ_pack/sketch.html
Code: https://github.com/quillaja/p5-sketches/tree/master/circ_pack


quillaja:
FYI, I made a number of improvement to the sketch's user interface since the pull request. This allows the user to alter some parameters of the algorithm and view a message box with help about controls, etc.

PR #616, opened 2018-06-04
Update 105-polynomial-regression-tfjs.md

MkLahane:
My polynomial regression version with n-degree polynomial interface.


shiffman:
yay, thank you!

PR #626, opened 2018-06-16
Converted CC_014 to processing

TheLastDestroyer:
https://github.com/CodingTrain/website/issues/267

Uses mouse scroll wheel instead of a slider to change angle
capture


nilsweb:
Thank you @TheLastDestroyer! 🎉

PR #645, opened 2018-06-24
Add port to p5.js of coding challenge 107 (sandpiles)

edorfaus:
This PR first adds a straight port (only the required changes) of the Processing code, and then adds some changes on top of that to make it run faster (since it was being very slow - less than 0.4 fps on my PC).

I tried to keep the final result close to the original Processing code, and kept the changes in a separate commit so people can see what I changed (and hopefully why) if they want to.


nilsweb:
Thanks a lot!

PR #648, opened 2018-06-28
Added a community contribution.

addowhite:
Added a link to a C++ implementation of the Barnsley Fern algorithm which renders the fern in a console with unicode characters.
https://www.youtube.com/watch?v=3Q5tAWkAEBc


addowhite:
Yeah that threw me a bit. Thanks for pointing me in the right direction!

PR #651, opened 2018-06-29
Update 003-snake-game-p5.md

MkLahane:
I created a snake game inspired by  the snake vs blocks game on play store.


meiamsome:
Thanks for your contribution, I patched up your PR a bit to match our style requirements (so the files don't become too different in terms of spacing), but nothing too major!

PR #669, opened 2018-07-16
Session 7 - Videos 1 & 2

nonas-hunter:
This pull request adds the code for the first two videos in the Session 7 - Intelligence and Learning playlist.

The code in this pull request is EXACTLY what is seen at the end of each video.

However - there is one small problem:
The code written in these videos is for data collection and DOES NOT include the actual classifier.
This also means that it uses the API Key which is now unusable due to permission restrictions set on the database. Therefore, I am not entirely sure if this code should be included in the repo ( this is a question for @shiffman ).

I will continue doing this for the rest of the video series. - Hope this helps!


shiffman:
Yes, I think it's ok if the code doesn't work with the data as the idea is for viewers to connect to their own data. The important thing is to have a reference to the code.

shiffman:
Thank you @nonas-hunter!

PR #670, opened 2018-07-16
Session 7 - Video 3 & 4

nonas-hunter:
I added the code for videos 3 and 4 which covers visualizing the data. The code works perfectly as shown in the video :)

I did not do any extra formatting or styling so final product looks the exact same way it does at the end of each video.

I also added author comments to the beginning of the code for videos 1 & 2.


shiffman:
This is amazing, thank you!

PR #672, opened 2018-07-18
Adding Code for Session 7 Video 7 and Edits for Video 6

nonas-hunter:
I double checked the video and it turns out the code in the video 6 folder actually matched the end of video 7. I made some very minor changes so that the code matched the end of 6 and made a new folder with the code for video 7.

We're half way there!!!! :)


shiffman:
Amazing, thank you!

PR #676, opened 2018-07-25
Cc 110 processing

TheLastDestroyer:
Ported the first part of CC 110 to processing.

Scaling changed slightly to improve performance and now saves an image on mouseclick


CodingChallenges/CC_110_recaman_processing/CC_110_recaman_processing.pde
```diff
@@ -0,0 +1,69 @@
...
+// Code for https://youtu.be/DhFZfzOvNTU
+// Processing port by Max (https://github.com/TheLastDestroyer)
+
+import java.util.Map;
+
+int current;
+int step;
+float scl;
+float oldscl = 1;
+HashMap<Integer, Boolean> numberline = new HashMap<Integer, Boolean>();
```
shiffman: How about trying Processing's `IntDict` for this instead and using `0` for false and `1` for true?

TheLastDestroyer: I didn’t know there was such a thing, I’ll have a look

TheLastDestroyer: (I’m a python guy really :) )

TheLastDestroyer: Unless I’m looking at the wrong thing IntDict looks like a string to int dictionary which I don’t Think is the best way to do this. Could cast the integer to a string for the key and then store 1 or 0 (or just 1 as I don’t store into we haven’t seen) but unless there is some technical reason this is better (processing is not my main language) I don’t see the need.
[docs](https://processing.org/reference/IntDict.html)

An intList might be the way to go though, has a handy hasValue() function as well

shiffman: Ah, right, yes, let's use `IntList`!

TheLastDestroyer:
that should be it (I think)

shiffman:
Great!

PR #677, opened 2018-07-26
Recamán's Sequence - Part 1 animated arc

jacekku:
Hello, I've modified the code so that the arc is animated instead of appearing all at once,
in my codepen the time between new numbers is set at 1s (1000ms).
Hopefully I modified the file correctly :)

link to working example: https://codepen.io/jacekWyroba/full/ajyaJE/


shiffman:
Hey, the animation looks great, would love to merge and show on Friday's live stream! Seems like there is a syntax error in the markup still, can you double check?

shiffman:
Yay!

PR #678, opened 2018-07-27
Adding useful features for educational purposes

marshalhayes:
This pull request adds a few features which may be useful for educational purposes. These features include:
- Generate data and fit line soon after page load
- Button to print data `x` and `y` values
- Show line equation `y = ${m}x + ${b}` when printing data

In addition to the above, I am also using `tf.train.adam(0.09)` with `x, y` mapping between `[-1, 1]`. I found that seeing the line react when adding data points can be entertaining and interesting, at least.

I plan to add UI elements to change the learning rate and optimizer soon, if time permits. 

Meanwhile, you can see a working version at [marshal.cool/js-linear-regression](https://marshal.cool/js-linear-regression).

### DIFF
--- a/CodingChallenges/CC_104_tf_linear_regression/index.html
+++ b/CodingChallenges/CC_104_tf_linear_regression/index.html
@@ -12,7 +12,7 @@
   <body>
     <div id="main">
       <header>
-        <a href="https://marshal.cool">< marshal.cool</a>
+        <a href="https://marshal.cool">marshal.cool</a>
         <h1 style="line-height: 0;">js-linear-regression</h1>
         <h3>interactive linear regression</h3>
       </header>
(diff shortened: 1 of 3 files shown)


marshalhayes:
I've added a few features such as a slider to change learning rate and a dropdown to change which optimizer is used. Let me know if you want me to submit another PR!

shiffman:
Hi @marshalhayes! This is fantastic work! 🎉  However, with the exception of minor style tweaks and bug fixes, I am trying to keep the code here "as is" so that it matches the video tutorials exactly. I think the work you are doing would make more sense as a "community contribution" that would like out from the challenge page itself? More info is on the wiki. 

https://github.com/CodingTrain/website/wiki

Does this make sense to you?

marshalhayes:
Thanks! I agree that this is a good idea.

PR #691, opened 2018-08-17
Added md files for CC112,113 and 114

blinkityblank:


shiffman:
Any idea why these pages aren't showing up? Or is it just me not seeing them? @nilsweb @meiamsome ?

meiamsome:
@shiffman @nilsweb I think it's because they were added before their release dates, and the index page hasn't been regenerated since. Regenerating it locally shows them on the index pages for me.

nilsweb:
Of course! I totally didn't look at the dates. The pages got generated but the sorting mechanism filters out future videos. I don't know if you can trigger a Jekyll rebuild on GitHub directly but as soon as the next change has been made to the repository, it should show up correctly. Thanks, @meiamsome!

shiffman:
Ah, makes sense, thank you! I will make some small edits and that should do it!

Issue #694, opened 2018-08-21
Rotation Matrix incorrect for 3D rendering Coding Challenge


mustafaquraish:
In CC_112_3D_Rendering.pde the following rotation matrix is defined

```
  float[][] rotationY = {
    { cos(angle), 0, -sin(angle)},
    { 0, 1, 0},
    { sin(angle), 0, cos(angle)}
  };
```
It should however be (noticed the signs)

```
  float[][] rotationY = {
    { cos(angle), 0, sin(angle)},
    { 0, 1, 0},
    { -sin(angle), 0, cos(angle)}
  };
```

Otherwise, the rotation happens along a different direction. [Wikipedia](https://en.wikipedia.org/wiki/Rotation_matrix)

shiffman:
Thank you for the fix!

PR #697, opened 2018-08-23
Fixed rotationX and rotationY

mustafaquraish:
Last row for rotationX had an extra 0.
rotationY was wrong. Issue #694


shiffman:
Thank you!

PR #707, opened 2018-08-30
Added my implemention of the bubblesort visualizer

ItsWasteD:
I created this with P5.js in like 15 minutes, so ist not very advanced but it fulfills ist purpose :)


nilsweb:
Your YAML was valid but we decided to not allow trailing spaces just to keep everything clean. That's why it was rejected. Maybe we should put that somewhere on the wiki page... 🤔

PR #713, opened 2018-09-06
added my example

tslytsly:
wrote it in p5.js on the web editor.
Managed to make it run faster as a Bubblesort, doesn't look like Selection sort anymore.


shiffman:
nice one!

PR #722, opened 2018-09-28
Switch from let to const

sleepy-maker:
using const might be better.


sleepy-maker:
@shiffman I see. That makes sense totally.

nilsweb:
I quickly added a short note in the file!

PR #730, opened 2018-10-02
fixing testing issue?

shiffman:
As per: https://github.com/jsdom/jsdom/issues/2304


shiffman:
This fixed the test but there's an error in a markdown file so I'm also force merging this!

PR #732, opened 2018-10-02
Fixed url for one of the videos

blinkityblank:


nilsweb:
I think internal links need to start with a leading `/`?

blinkityblank:
Yes I believe you're right. I fixed it in the following commit.

nilsweb:
Now everything seems to be working again! 🎉

PR #735, opened 2018-10-06
Added hover effect on .video-card

joerez:


gruselhaus:
@joerez Thanks for that! I've corrected the syntax so that it can be successfully compiled. May I merge from your side?

## Preview:

jan-06-2019 16-38-57

gruselhaus:
@nilsweb I added a restriction there. This should fix the problem.

joerez:
Glad this ended up working out.

gruselhaus:
@shiffman What do you think about it?

PR #751, opened 2018-10-12
Fixed Gemfile Vulnerabilities

nilsweb:
This *should* fix the `Gemfile`/`Gemfile.lock` vulnerabilities as shown on the GitHub page.
I'm now quite sure if it got everything because GitHub doesn't show the message on my end for some reason. Maybe @shiffman could check?

It also addresses the issue that for some reason bundler (the package manager used to install Jekyll and its dependencies) couldn't finish installing after running `bundle install` due to some weird "native-extension-building" errors.


shiffman:
I think this is good so I'm going to merge and see!

PR #757, opened 2018-10-14
 Update the CodingCallenges/README.md

gruselhaus:


shiffman:
I think we should remove them!  If there is a community contribution that we don't have on a challenge page great if we can add it?

PR #761, opened 2018-10-15
Added my example

tslytsly:


shiffman:
Amazing! You should share this with Tom Scott! (Maybe on twitter?)

tslytsly:
Thanks, maybe I will...

PR #778, opened 2018-10-17
Added a community contribution o/

TheTastefulToastie:
Wrote a seven segment display visualisation and tried to use a cascading switch statement to exploit that some numbers can be drawn as a combination of other numbers.


nilsweb:
Thanks for sharing! 👍

Issue #734, opened 2018-10-06
New "Coding Train" banner


shiffman:
There are new designs that could be incorporated into the banner etc. Would love help with this!

All the design elements are here:

https://drive.google.com/drive/folders/0B_AnqhTfj4qjTlBnNm1vcktPWVE?usp=sharing

(New wallpapers should be added and a way for people to download wallpapers re: #474?)

bagshik:
Hey, I can help with this. Can you elaborate what needs to be done? :)

nilsweb:
Sure! The current banner on the [website](https://thecodingrain.com) is still the old version of the banner. The only "problem" is that the banner might not be long enough so that it also looks good on e.g. ultra-wide screens. To get the current banner working I took the Illustrator file and extended the background and then re-exported it. @TurqDevDesign also did some JPEG-magic to make it load faster when using slower connections. Unfortunately I don't know exactly how he did it but maybe I can get it to work somehow and post the new banner here so you can implement it?

bagshik:
Sounds good, I'll wait for your respone 👌

TurqDevDesign:
@nilsweb I saved the file with Progressive refine. When you save a file in illustrator you should be able to 'Save for Web', I believe the shortcut is CMD + Option + Shift + S on mac (I'm not sure if it has changed, I've been using Affinity designer to edit vectors lately 😛) The Save For Web interface should look something like this: https://i.ytimg.com/vi/sUJvTv90ybs/maxresdefault.jpg , Just tick that "Progressive" box in the top right corner.

nilsweb:
Thanks for the explanation @TurqDevDesign! I think I got it right. 😅

Here's the new banner:
![new_header](https://user-images.githubusercontent.com/12577464/46906199-1a1f2b00-cf00-11e8-870b-e68ef361be7f.jpg)

@bagshik You just have to replace it now!

PR #782, opened 2018-10-18
Added new banner

bagshik:
### Issue #734


nilsweb:
Thanks @bagshik!

Closes #734

PR #805, opened 2018-10-21
Create README.md

MericLuc:
Added my own version of your Seven-segment display challenge! =)

This is not very different from yours, except i used a OOP design. 

You can have a look of a screenshot here:
https://github.com/MericLuc/CC117_SevenSegment/blob/master/example.png


nilsweb:
Thank you for sharing this! 👍

Instead of creating a new `.md` file, you can just use the one that is already there. That way, it also shows up on the website. Here's the link to the file you want to edit: https://github.com/CodingTrain/website/edit/master/_CodingChallenges/117-seven-segment.md

PR #813, opened 2018-10-23
Community contribution to #119: Ascii to binary converter

doxanthropos:
Based on the binary visualization from Challenge 119, I build an ASCII to binary visualization: https://doxanthropos.github.io/sketchbook/2018102301/


shiffman:
Nice job!!

PR #822, opened 2018-10-27
Fixed View/Download Code and Live-Example

### LINKED ISSUE #784: The issue out the Live stream right now!!!
knipsel

This is the screenshot out the stream

nilsweb:
This fixes the problems mentioned in #784. The "new way" detects P5 and Processing code automatically and displays the links to view/download the code on the website. If the `web_editor` property is set, it will also be used.

Since it can be very confusing to describe what it does with words, I've made a little diagram:
ct - sketches

I've also migrated all community contributions that I've found in the old `README.md`s to the Markdown files.

The only thing that's kind of "broken" at the moment is the ability to view/download code for tutorials and courses directly from the website. Those have to be migrated to the new format like I did with all the coding challenges. Maybe we can get some members of the community to help with that?

---
Fixes #784


shiffman:
This is amazing, I am so excited about it! After merging it may be worth revisiting the "tutorials" vs "courses" question. Is this a meaningful distinction? I'll open a new issue to discuss.

nilsweb:
Yes, definitely! Especially when moving away from Jekyll at some point.

shiffman:
Ok, merging, excited to try it out!

PR #858, opened 2018-11-20
Add addtional info for screen readers

### LINKED ISSUE #837: Multiple of the same buttons - accessibility issue
Some pages like https://thecodingtrain.com/CodingChallenges/ have multiple buttons with the same text. "Watch on YT" for example. While this works fine it's problematic for accessibility. 

One off the most used features of a screen reader is getting a list with all the links on a page.
When users are presented with a bunch of the same text, they are forced to go through the whole page to get a clearer picture.

What I suggest is to add a more detailed description of the link in a span that puts it offscreen.

`.sr-only {
  position: absolute;
  width: 1px;
  height: 1px;
  padding: 0;
  margin: -1px;
  overflow: hidden;
  clip: rect(0,0,0,0);
  border: 0;
}`

`<a href="#">
   Watch on YT 
   <span class="sr-only">: Coding challenge #121 - Logo Interpreter</span>
</a>`

bjorvack:
fixes #837

Added aria-label attributes to video cards with a hidden span as fallback

### DIFF
--- a/_jekyll/includes/3-modules/video-card.html
+++ b/_jekyll/includes/3-modules/video-card.html
@@ -12,9 +12,14 @@
 
   {% comment %} Use right link based on the video's date. {% endcomment %}
   {% if include.video.date <= site.time %}
-    <a href="{{ include.video.url | relative_url }}">
+    <a href="{{ include.video.url | relative_url }}"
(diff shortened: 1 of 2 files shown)


nilsweb:
Thanks for sharing this!
Alright, so I think we can merge then? Is that OK with you, @shiffman?

PR #875, opened 2018-12-02
Added Node.js as a code variation

gruselhaus:
Moin (a typical north german greeting),

I think we should add Node.js as a code variation. This describes some challenges in more detail. For example, the Mastodon Bot Challenge is currently categorized under "JavaScript". "Node.js" would be more accurate.

@shiffman What do you think?
@nilsweb FYI


shiffman:
Yes, I agree! I'm also willing pay for a server and have some sort of process where the node projects are deployed and can be run if that makes sense!

_config.yml
```diff
@@ -40,6 +40,8 @@ code_variations:
     folder: "Processing"
   - name: "JavaScript"
     folder: "JavaScript"
+  - name: "NODE.JS"
```
shiffman: Is there a reason why this is all capital?
```suggestion
  - name: "Node.js"
```

gruselhaus: There'll no difference. The style will capitalize it anyway.

nilsweb:
Thank you! :tada:

PR #878, opened 2018-12-04
Contribution from John Pisello (John30013)

John30013:
My version of the Chaos Game, with sliders to control various parameters, and leveraging vector properties to color the pixels.


shiffman:
So cool!

PR #879, opened 2018-12-04
implemented Quicksort for pixel sorting

thetornadotitan:
Implemented Quicksort for pixel sorting.

I created a directory for the Quicksort inside of the pixel sorting coding challenge folder.


gruselhaus:
@thetornadotitan Thank you for this submission. Can you submit this as a community contribution instead (see: https://github.com/CodingTrain/website/wiki/Community-Contributions-Guide) So it would be clearer that this came from you and not from our train driver @shiffman ? 🚂

PR #896, opened 2018-12-11
QuadTree community contribution

TheTastefulToastie:
Added QuadTree optimization to the p5.js web editor sketch so we can crank up the size of those flocks.


shiffman:
Oh, this is awesome thank you!

PR #898, opened 2018-12-12
Sharing colorful regular polygons flocking

jcponce:
Colorful regular polygons flocking

[https://editor.p5js.org/jcponce/sketches/BkCDZSR1V](url)


shiffman:
Thanks for this!

PR #124, opened 2017-03-24
added README.md to CC_67_2_binary_tree_viz

OtterTricks:
also added a link to my community variant in the README.md

PR #900, opened 2018-12-14
Contribution to coding challenge #124 : Evolution Aquarium

anuraghazra:
My Contribution to coding challenge #124.
added small Creatures, Predators, Avoiders, Food, Poison etc.

### DIFF
--- a/_CodingChallenges/124-flocking-boids.md
+++ b/_CodingChallenges/124-flocking-boids.md
@@ -39,6 +39,12 @@ contributions:
       name: "Sebastien Raynaud"
       url: "https://twitter.com/Chopokopx"
     url: "https://editor.p5js.org/SebastienR/sketches/ryDc4qaJE"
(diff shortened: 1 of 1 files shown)


shiffman:
This is so cool! Do you have a write-up or anything that explains all the different behaviors etc.? Great work! ❤️

anuraghazra:
Actually I don't have a write-up for now but the code is very simple.
If the creatures eat poison their health goes down, and if they eat food their health goes up (inverse for Predators).
I also added a **behavior();** function which handles *food(like)* and *poison(dislike)* perception.

All the agents have a **defineFear();** function which handles the behavior.

And there are Male(green) and Female(pink) creatures, and they grow big as they eat more food and when they became adults they reproduce(); (if they are close enough) over time.

And for the “Predators” they also use **defineFear();** but inversely.

PR #908, opened 2018-12-19
Contributing my 3D flocking simulator (25.600 particles)

EthanHermsey:
It uses p5.js, three.js and a custom made feedback-buffer-object engine to calculate the accelerations, velocities and positions of 25.600 particles (160*160) using the GPU.


shiffman:
whoa!

PR #913, opened 2018-12-20
Removing the rest of the path array

nijan18:
In the livestream a path array was made to draw the path of the point.
This, however, was not included in the final video, therefore some unnecessary code for the path remained in the code.


shiffman:
Thank you!!

PR #914, opened 2018-12-21
Added target="_blank" to code-variation-dropdown.html

gruselhaus:
@nilsweb What do you think about that? Was that thought?


shiffman:
Great fix!

PR #930, opened 2018-12-22
Transcribe Hypercube to javascript.

FriskyDev:
Would love to create a better matrix math library or use one of the existing ones. The .array() function, and set() operators of the p5.Vector would seem to make this easier, but the arrays are one dimensional, so it really feels like there needs to be a matrix class, with corresponding conversions. (and I am thinking out loud: inside p5.js)

I say this because I used similar, **but slightly different code** for another one of the samples. Then I used that one to create **a slightly different one** I hooked up to the Barnsley Fern example (which was doing the algebraic equivalent).

### DIFF
--- a/CodingChallenges/CC_113_Hypercube/P5/P4Vector.js
+++ b/CodingChallenges/CC_113_Hypercube/P5/P4Vector.js
@@ -1,0 +1,27 @@
+// Daniel Shiffman
+// http://youtube.com/thecodingtrain
+// http://codingtra.in
+// Javascript transcription: Chuck England
+
+// Coding Challenge #113: 4D Hypercube
+// https://youtu.be/XE3YDVdQSPo
+
+// Matrix Multiplication
+// https://youtu.be/tzsgS19RRc8
+
+
+class P4Vector {
+    constructor(x, y, z, w) {
+        this.x = x || 0;
+        this.y = y || 0;
+        this.z = z || 0;
+        this.w = w || 0;
+    }
+
+    mult(f) {
+        this.x *= f;
+        this.y *= f;
+        this.z *= f;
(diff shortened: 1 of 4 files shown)


FriskyDev:
There error is from some new check?

FriskyDev:
🥇 thanks @gruselhaus

CodingChallenges/CC_113_Hypercube/P5/matrix.js
```diff
@@ -0,0 +1,99 @@
+// Javascript transcription: Chuck England
```
shiffman:
```suggestion
// JavaScript transcription: Chuck England
```

CodingChallenges/CC_113_Hypercube/P5/sketch.js
```diff
@@ -0,0 +1,109 @@
+// Daniel Shiffman
+// http://youtube.com/thecodingtrain
+// http://codingtra.in
+// Javascript transcription: Chuck England
```
shiffman:
```suggestion
// JavaScript transcription: Chuck England
```

CodingChallenges/CC_113_Hypercube/P5/P4Vector.js
```diff
@@ -0,0 +1,27 @@
+// Daniel Shiffman
+// http://youtube.com/thecodingtrain
+// http://codingtra.in
+// Javascript transcription: Chuck England
```
gruselhaus:
```suggestion
// JavaScript transcription: Chuck England
```
Here is the last one. Then we can merge :D

gruselhaus:
@FriskyDev Thank you ❤️

PR #123, opened 2017-03-24
Master

tobylockley:
Added readme with a link to my version with rainbow colors and zooming.

PR #932, opened 2018-12-22
Enforce Web Editor existence, user & mapping to repositories

meiamsome:
Makes tests check that a specified `web_editor` field exists, is on the right account and has an optional 1-to-1 mapping to a repository

### DIFF
--- a/unit_testing/package-lock.json
+++ b/unit_testing/package-lock.json
@@ -535,7 +535,7 @@
         },
         "chalk": {
           "version": "1.1.3",
-          "resolved": "http://registry.npmjs.org/chalk/-/chalk-1.1.3.tgz",
+          "resolved": "https://registry.npmjs.org/chalk/-/chalk-1.1.3.tgz",
           "integrity": "sha1-qBFcVeSnAv5NFQq9OHKCKn4J/Jg=",
           "dev": true,
           "requires": {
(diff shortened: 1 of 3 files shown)


meiamsome:
In particular, do you think we should be enforcing the titles like this, or should we be forcing the title update on push to master. My thoughts for testing title is so that someone doesn't accidentally add the `web_editor` field from a different challenge. Otherwise, I could add a check that the `web_editor` fields are unique?

nilsweb:
Both are valid points. The title _should_ match so people don't get confused and think that they accidentally clicked on the wrong link. Having unique `web_editor` fields could be problematic though, since there would be no way for having the same `web_editor` on multi-part challenges.

meiamsome:
@nilsweb yeah, unset values are just ignored, sadly the output doesn't make that clear, but it is what's happening.

As per test performance, we can rework it using `test.concurrent` later I think that would make them all run parallel if it becomes slow, but for now it seems fast enough.

The issue with unique `web_editor` fields is the same as enforcing the title, though - with this suggestion, having a part 1 and a part 2 which naturally have separate titles have the same `web_editor` would throw a mismatched title error

nilsweb:
Perfect! :+1:

You're absolutely right! I haven't thought of that... 🙃 Then, I would not check for consistent titles or unique `web_editor` values.

Maybe... Is it possible to check for unique `web_editor`s and ignore multiple occurrences within the "main" video number? So it would be ignored for `#123.1` and `#123.2` but not for `#124`? Do you know what I mean? :thinking:

meiamsome:
Ok, so I removed the title requirement, and added a one-to-one mapping between repository and web_editor, so that the mapping between them is kept the same across files.  We can add a `title` check later for ones that are only used in one file perhaps? Or just leave it out, it's not too important, most user error situations will be caught by this.

PR #943, opened 2018-12-24
Transcribe Logo_2 to Processing

FriskyDev:
I found some consistency issues with Logo_1 I did. I think the Command interface I put in that one should match this one where its called Action, since a Command was added.

I also did not like the argument passing in this one, which Javascript lets you dance past the issue. 

Will wait for feedback to update both. ;)


FriskyDev:
The revert was because I accidentally started from Logo_1 branch, and this makes the changes here just the Logo_2 changes.

PR #961, opened 2018-12-31
Fix 128

ciscou:
We should reset the `pen` to `"down"` before starting a new sketch, otherwise we won't draw the first stroke of every sketch.

### DIFF
--- a/CodingChallenges/CC_128_SketchRNN_Snowflakes/P5/sketch.js
+++ b/CodingChallenges/CC_128_SketchRNN_Snowflakes/P5/sketch.js
@@ -29,6 +29,7 @@ function draw() {
   if (strokePath != null) {
     let newX = x + strokePath.dx * 0.2;
     let newY = y + strokePath.dy * 0.2;
+    if(pen === "end") console.log("pen is end");
     if (pen == "down") {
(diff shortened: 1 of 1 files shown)


CodingChallenges/CC_128_SketchRNN_Snowflakes/P5/sketch.js
```diff
@@ -8,12 +8,11 @@ let model;
...
 
 let x, y;
-let pen = "down";
+let pen;
 
 function setup() {
   createCanvas(windowWidth, windowHeight);
-  x = random(-width / 2, width / 2);
-  y = random(-height / 2, height / 2);
+  setupNewSketch();
```
shiffman: Would you mind adding some comments explaining what's changed here from the video? If you prefer I can add them.

ciscou: @shiffman I've done so in the last commit. I'm not sure if that's what you wanted me to do, and also I'm not a native English speaker, so feel free to revert and add the comments yourself if you want to :)

shiffman:
Great, made some minor edits, thank you!!!

PR #974, opened 2019-01-04
Added contribution

norman521:
This is an object oriented implementation of fractal trees that creates random forests


gruselhaus:
Great idea and implementation ❤️

PR #976, opened 2019-01-04
Transcribing Pixel Sorting to Javascript.

FriskyDev:
I have done the work for both the non-animated and the animated version in this PR.

Please note that in Javascript, this code **runs abysmally slow**. The animated version at least shows the progress as it works.

It would not hurt my feelings at all if you wanted to just not accept this PR.


FriskyDev:
I had to move the animated one into the P5 folder. There is a sub-folder with the work for the non-animated version in it.

FriskyDev:
> I think this is useful for people who are looking! We could consider reducing the image to 100x100?

💯 
I shrunk the image to 100x100 as you suggested, and it seems reasonably fast enough. I removed the 400x400 image (although its available in the processing folder). I adjusted the code slightly for the new image sizes.

PR #980, opened 2019-01-05
Transcribe "tweets by month" to Processing.

FriskyDev:
This was an interesting one. There are several language style differences.

I have used maps, array lists, and set collections here to keep the code simplified. The code can always use purely arrays. However, in this particular case, it feels like you really need the collections to match the equivalent object/property abilities found in Javascript. In addition, there would need to be additional routines written to perform lookups, sorting, searching, etc. Glad to add these, but it takes the sample farther from the original.

In order to store the totals/word counts, I created a class, that better mimics the Javascript object that was being created.

The string being used to sort and store the tweets by date worked in Javascript, but only based on the language semantics. In the Processing version, I created the string in reverse order (year/month) and then used sorting (instead of reversing) to acquire the same result.

The date parsing in Java is also more complex. To parse, you have to create a DateFormat instance that matches the format of the dates, and then call its parse function. Since this can throw, it all has to be wrapped in a try/catch block.

The drawing code is pretty much identical.

The HTML element was drawn at the top of the screen.


CodingChallenges/CC_048_TweetsByMonth/Processing/CC_048_TweetsByMonth/CC_048_TweetsByMonth.pde
```diff
@@ -0,0 +1,137 @@
...
+// YouTube video tutorial: https://youtu.be/UrznYJltZrU
+// Javascript transcription: Chuck England
+
+import java.util.*;
+import java.text.*;
+
+// Class to hold the monthly totals and word counts.
+class Count { 
+  int total;
+  Map<String, Integer> words;
```
shiffman: Would it be possible to use Processing's native `IntDict` class? 
https://processing.org/reference/IntDict.html

FriskyDev: I think you pointed one of these out before. I will switch this one out and try to remember it for the future. :)

PR #981, opened 2019-01-05
Transcribe A-star to Processing.

FriskyDev:
Several dynamic arrays here, so I used the ArrayList and its methods.

Can re-write for pure arrays on request.

### DIFF
--- a/CodingChallenges/CC_051_astar/Processing/CC_051_astar/CC_051_astar.pde
+++ b/CodingChallenges/CC_051_astar/Processing/CC_051_astar/CC_051_astar.pde
@@ -1,0 +1,169 @@
+// Daniel Shiffman
+// http://codingtra.in
+// http://patreon.com/codingtrain
+// Part 1: https://youtu.be/aKYlikFAV4k
+// Part 2: https://youtu.be/EaZxUCWAjb0
+// Part 3: https://youtu.be/jwRT4PCT6RU
+// Processing transcription: Chuck England
+
+// An educated guess of how far it is between two points
+float heuristic(Spot a, Spot b) {
+  float d = dist(a.i, a.j, b.i, b.j);
+  // float d = abs(a.i - b.i) + abs(a.j - b.j);
+  return d;
+}
+
+// How many columns and rows?
+int cols = 50;
+int rows = 50;
+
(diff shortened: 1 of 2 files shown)


CodingChallenges/CC_051_astar/Processing/CC_051_astar/Spot.pde
```diff
@@ -0,0 +1,88 @@
...
+      wall = true;
+    }
+  }
+  
+  // Display me
+  void show() {
+    if (wall) {
+      fill(0);
+      noStroke();
+      ellipse((float)i * w + w / 2.0, (float)j * h + h / 2.0, w / 2.0, h / 2.0);
```
shiffman: Is the casting here necessary?

CodingChallenges/CC_051_astar/Processing/CC_051_astar/CC_051_astar.pde
```diff
@@ -0,0 +1,169 @@
...
+
+// The road taken
+List<Spot> path = new ArrayList<Spot>();
+
+void setup() {
+  size(400, 400);
+  println("A*");
+
+  // Grid cell size
+  w = width / (float)cols;
```
shiffman:
```suggestion
  w = float(width) / cols;
```

Processing has a `float` method which is admittedly a little weird, but makes the code a bit easier on the eyes for beginners.

CodingChallenges/CC_051_astar/Processing/CC_051_astar/CC_051_astar.pde
```diff
@@ -0,0 +1,169 @@
...
+  h = height / (float)rows;
```
shiffman:
```suggestion
  h = float(height) / rows;
```

CodingChallenges/CC_051_astar/Processing/CC_051_astar/Spot.pde
FriskyDev: Hmm, I was actually hunting a bug when I did it. I think not, but I'll double check.
The issue was that the h variable in the class was shadowing the global one. So I changed the name of the variable. Removing the casts.

PR #982, opened 2019-01-05
Transcribe "random walk" to Processing.

FriskyDev:


CodingChallenges/CC_052_random_walk/Processing/CC_052_random_walk/CC_052_random_walk.pde
```diff
@@ -0,0 +1,38 @@
...
+  y = 200;
+  background(51);
+}
+
+void draw() {
+  stroke(255, 100);
+  strokeWeight(2);
+  point(x, y);
+
+  int r = floor(random(4));
```
shiffman:
```suggestion
  int r = int(random(4));
```

Tiniest of points but Processing has an `int()` method to convert to integer which is more typical of examples (I think?)

PR #983, opened 2019-01-05
Transcribe "random walk levy" to Processing.

FriskyDev:


CodingChallenges/CC_053_random_walk_levy/Processing/CC_053_random_walk_levy/CC_053_random_walk_levy.pde
```diff
@@ -0,0 +1,38 @@
...
+// int y;
+PVector pos;
+PVector prev;
+
+void setup() {
+  size(400, 400);
+  background(51);
+  pos = new PVector(200, 200);
+  prev = pos.copy();
+  println(pos);
```
shiffman:
```suggestion
  println(pos.x, pos.y);
```

I would probably remove this line but if we were to keep it makes more sense to print the values in Processing as Java will just output the memory address of the object.

PR #992, opened 2019-01-08
Renaming Processing file folder, as "." is not allowed in name.

FriskyDev:
Apparently, the period is not allowed in the filename, which you find when you attempt to open in processing.

Replaced the period (".") with and underscore ("_"). Have to do this to the .pde name and parent folder.


gruselhaus:
@FriskyDev Looks like all multi-part processing challenges have to be renamed.Could you please take a look?

FriskyDev:
> @FriskyDev Looks like all multi-part processing challenges have to be renamed. Could you please take a look?

@gruselhaus Yes. I wanted to make sure I had it right and was not missing something. I had to do this for one of the Processing transcriptions I did. But, I will go through and look for any others and make another single PR with the changes.

gruselhaus:
@FriskyDev Looks good for me. Let's merge ❤️ Thanks!

PR #998, opened 2019-01-11
Transcribe "binary tree viz" to Processing.

FriskyDev:
The tree representation overdraws itself in a lot of instances. It is a very complicated spacial problem as the space required potentially doubles with each row.

This version is a straight port of the original code.

I saw this interesting article on drawing binary trees: http://llimllib.github.io/pymag-trees/
Perhaps that is an episode? :)

There are many others online. Some of the more interesting ones show insertions with tree rotations of red-black/AVL trees, and stuff like that.

### DIFF
--- a/CodingChallenges/CC_065.2_binary_tree_viz/Processing/CC_065_2_binary_tree_viz/CC_065_2_binary_tree_viz.pde
+++ b/CodingChallenges/CC_065.2_binary_tree_viz/Processing/CC_065_2_binary_tree_viz/CC_065_2_binary_tree_viz.pde
@@ -14,7 +14,7 @@ void setup() {
   background(51);
   tree = new Tree();
   for (int i = 0; i < 100; i++) {
-    tree.addValue(floor(random(0, 100)));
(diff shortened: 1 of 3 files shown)


CodingChallenges/CC_065.2_binary_tree_viz/Processing/CC_065_2_binary_tree_viz/CC_065_2_binary_tree_viz.pde
```diff
@@ -0,0 +1,28 @@
...
+// Part 2: https://youtu.be/KFEvF_ymuzY
+
+Tree tree;
+
+void setup() {
+  size(600, 400);
+  background(51);
+  tree = new Tree();
+  for (int i = 0; i < 100; i++) {
+    tree.addValue(floor(random(0, 100)));
```
shiffman:
```suggestion
    tree.addValue(int(random(0, 100)));
```

shiffman:
Thanks for that reference, great article about rendering trees!

Issue #904, opened 2018-12-15
"Nature of Code" Link doesn't work on mobile devices


gruselhaus:
![bildschirmfoto 2018-12-15 um 16 09 04](https://user-images.githubusercontent.com/33380107/50044469-0caf2a00-0084-11e9-8767-c2b906a33990.png)

nilsweb:
This looks like something I may have seen before, then fixed it but somehow it still happens. For some reason, certain portions of the screen are completely unresponsive when on mobile. Maybe it's the menu that overlaps elements even when closed? I haven't got the time to dig deeper into this yet, maybe someone else wants to explore it further?

baddox:
The same is true for the purple hover menus like “view code” and “download code” on pages like https://thecodingtrain.com/CodingChallenges/124-flocking-boids

ramsaysewell:
Seems to be the navigation container is overlaying the three buttons. When adding "display: none" to .navigation, the footer is clickable. Maybe a solution could be to "display: none" on mobile screens until the .nav-toggle is clicked.

PR #1002, opened 2019-01-12
Link doesn't work on mobile devices -fix

Fraasi:
The culprit is the nav element just under .navigation  as talked about in #904  
This changes it's width to 0% unless the nav-toggle is checked, which hopefully solves the problem.
**Note:** I have no way of building and testing jekyll site, so hopefully I got the sass-code right.

### DIFF
--- a/assets/css/2-base/_navigation.sass
+++ b/assets/css/2-base/_navigation.sass
@@ -36,7 +36,7 @@
     width: auto
 
   nav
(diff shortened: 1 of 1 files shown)


gruselhaus:
@Fraasi Thanks for this! I'll run some test locally and report.

shiffman:
Amazing, thank you all! 💯

PR #1036, opened 2019-01-29
web editor code for challenges 52 and 53

shiffman:
This adds the web editor for challenges 52 and 53.


shiffman:
(I wonder why all these previous commits show up in this pull request?) Squashing and merging anyway!

PR #1040, opened 2019-01-30
added a grey wallpaper

DreamerPlays:
Hey,
I thought it would be a great idea to have a grey wallpaper that fits dark themes. With a bit of editing, I finally made something that fits my setup. I added a bit of color so it not 100% greyscale. I hope you enjoy it.

Dreamer


DreamerPlays:
Good addition, forgot that. Glad to help 😉

DreamerPlays:
It would be lovely if I had that. I am planning on a white themed image but I am thinking about how I do it with high quality. I would be very thankful if you could send the illustrator files to my email: yt.dreamer.plays@gmail.com

Dreamer

PR #1095, opened 2019-02-12
Added CC132 - fluid simulation

blinkityblank:


shiffman:
Hmmm, this should pass now that I added the code. . . is a p5.js version required? @nilsweb @meiamsome?

shiffman:
i've checked this and I think it's ok going to force merge 😬

nilsweb:
Sorry, I wanted to look at it all day but unfortunately didn't got to it.
Anyways, looks good and should work! I can't see why the test fails however... 🤔 Maybe @meiamsome might be able to figure it out... 😬

meiamsome:
Tests are failing because @blinkityblank's master is not in line with current master and so was missing the folder, when in our master it is fine. The merge of master in the commit history here is from their master and not ours and so still missed the files.

Also related is that CircleCI continues to not provide the incredibly useful feature of running tests on the result of merging the code into the given branch (See for reference https://discuss.circleci.com/t/show-test-results-for-prospective-merge-of-a-github-pr/1662/24 and other places), which would have shown up as this test working correctly after the files had been added.

PR #1119, opened 2019-02-15
Cc 130 processing

TheLastDestroyer:
A port of the Fourier transform to processing, probably a bit less elegant than the version in part 3 but it works


gruselhaus:
Fixed in #1122

PR #1124, opened 2019-02-16
Innovative corner hit prediction

Turrett:
My program can understand when the dvd logo is going to hit the corner and reproduce a sound that can be timed to drop as soon as the logo hits the corner


gruselhaus:
❤️

PR #1144, opened 2019-02-19
P5 Version of CC134.2 & P5 Port of CC135

gruselhaus:


CodingChallenges/CC_135_GIF_Loop/P5/sketch.js
```diff
@@ -0,0 +1,40 @@
...
+  if (record) {
+    percent = float(counter) / totalFrames;
+  } else {
+    percent = float(counter % totalFrames) / totalFrames;
+  }
+  render(percent);
+  if (record) {
+    save("output/gif-" + nf(counter, 3) + ".png");
+    if (counter == totalFrames - 1) {
+      exit();
```
shiffman:
```suggestion
      noLoop();
```

`exit()` doesn's apply for p5. Also it might be worth investigating https://p5js.org/reference/#/p5/saveFrames? Eventually I'd like to make an example with https://github.com/wenheLI/p5.gif/!

gruselhaus: >Also it might be worth investigating https://p5js.org/reference/#/p5/saveFrames? Eventually I'd like to make an example with https://github.com/wenheLI/p5.gif/!

I'll take a look.

PR #1162, opened 2019-02-26
Added CC136.2 and added parts to CC136.1

blinkityblank:
@shiffman Needs the Repo when added and web editor id if it's different from part 1


shiffman:
Now in #1165, sorry I messed this up @blinkityblank!

PR #1166, opened 2019-02-26
 P5 Port of CC136.2

gruselhaus:


CodingChallenges/CC_136_Polar_Noise_Loop_2/P5/sketch.js
```diff
@@ -0,0 +1,45 @@
...
+function draw() {
+  let percent = 0;
+  if (record) {
+    percent = float(counter) / totalFrames;
+  } else {
+    percent = float(counter % totalFrames) / totalFrames;
+  }
+  render(percent);
+  if (record) {
+    saveFrame("output/gif-" + nf(counter, 3) + ".png");
```
shiffman: I wonder if the p5 ports should not actually save any files but just run the animation? Or does this perform ok? I don't think `saveFrame()` is a p5 function btw.

gruselhaus: > I wonder if the p5 ports should not actually save any files but just run the animation?

Yes, that makes sense.

PR #1171, opened 2019-03-04
my contribution to the codig challenge!!

alt-art:
i create a caracters noise loop, i guess
perlin noise caracters loop


gruselhaus:
wow, this is great ❤️ Thank you for this contribution!

PR #1178, opened 2019-03-05
add contribution flocking in 3D

tihawk:
This is flocking demonstrated using a different model, called the Vicsek Model, used in scientific research regarding flocking and swarming behaviour.
Demo: https://codepen.io/tihawk/full/vPXebG


gruselhaus:
❤️

PR #1199, opened 2019-03-14
Edit l.41 for digit count. Animation without using Pi

tht-jxny:
https://editor.p5js.org/tht-jxny/full/RG9tqIJqN Fullscreen version for PI Collision Animation without using PI. To change number of digits modify line 41 to a mass thats a power of 100. I've created this animation before the CodingTrain Video came out based on 3Blue1Brown's video so its neither OOP nor can it deal with more than 4 digits (at least not my pc :D). To try more just reduce the initial velocity.


gruselhaus:
❤️ Thank you !

PR #1202, opened 2019-03-15
Updating the p5.js references to 0.7.3

gruselhaus:


gruselhaus:
@meiamsome Could you double check the rules for the replacement? Four 👀 see more than two ;)

gruselhaus:
@shiffman Should we update the Coding Challenges only or should we update the Tutorials, Courses, ... as well?

gruselhaus:
@shiffman Should we update the Coding Challenges only or should we update the Tutorials, Courses, ... as well?

shiffman:
I'm good with updating everything!

gruselhaus:
Ok. I'm ready to go. Let's wait until @meiamsome reviewed and then merge 🚀

package.json
```diff
@@ -3,7 +3,9 @@
   "version": "0.0.1",
   "description": "Coding Train Website",
   "main": "none",
-  "dependencies": {},
+  "dependencies": {
+    "replace-in-file": "^3.4.4"
```
meiamsome: Perhaps this should be a dev dep instead of a dependency?

gruselhaus: Agreed!

p5Update.js
```diff
@@ -0,0 +1,91 @@
...
+(() => {
+  asyncForEach(options, async option => {
+    try {
+      const changes = await replace(option);
+      console.log(`${option.name}: `, changes.join(", "));
+    } catch (error) {
+      console.error("Error occurred:", error);
+    }
+  });
+})();
```
meiamsome: There should be a new line here ideally!

Any reason for the for each helper? You ought to be able to just use an async IIFE:
```javascript
(async () => {
  for (const option of options) {
    try {
      const changes = await replace(option);
      console.log(`${option.name}: `, changes.join(", "));
    } catch (error) {
      console.error("Error occurred:", error);
    }
  }
})();
```

gruselhaus: Oh, great. Thank you. I'm not that familiar with node.js.

prcoy:
How do I download the files for CC_064??? There is no zip file for them. Do have to download _individual_ files??

PR #1207, opened 2019-03-17
Added Contribution To CC46 - Asteroids (GyroDodge)

anuraghazra:
I used gyroscope sensor to control the player and added some sound effects too.
**GyroDodge is more fun to play on mobile devices.**

play : https://anuraghazra.github.io/GyroDodge/

### DIFF
--- a/_CodingChallenges/046.1-asteroids-part-1.md
+++ b/_CodingChallenges/046.1-asteroids-part-1.md
@@ -27,6 +27,12 @@ contributions:
       url: "https://github.com/davenewt"
     url: "https://davenewt.github.io/p5-asteroids/"
     source: "https://github.com/davenewt/p5-asteroids"
+  - title: "GyroDodge"
+    author:
+      name: "Anurag Hazra"
+      url: "https://github.com/anuraghazra"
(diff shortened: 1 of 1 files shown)


gruselhaus:
@anuraghazra Does it only run on android devices? On my iPhone it didn't work.

anuraghazra:
I don't know because I'm using Android device and I don't have access to an IPhone so i think I might have to consider Support for Safari's javascript deviceorientation api.

gruselhaus:
> Btw can you tell me which Safari version and IPhone you are using?

iPhone X, iOS: 12.1.4

anuraghazra:
Ohh okay, I think the support is not good. 

https://caniuse.com/#search=DeviceOrientation

gruselhaus:
Ok. Let's merge anyway ❤️

anuraghazra:
Okay thanks! i will try to solve the issue if possible.

anuraghazra:
Okay i think i can fix it. looks like IOS devices uses `ondevicemotion` rather than `ondeviceorientation` event.

https://stackoverflow.com/questions/4378435/how-to-access-accelerometer-gyroscope-data-from-javascript

PR #1213, opened 2019-03-18
Added contribution

GCDeveloper:
My contribution is an extension of the p5 code Shiffman used in the Coding Challenge video on calculating PI with Leibniz algorithm. Mine takes the average of the last 2 iterations to get a closer approximation to PI, much faster. It also displays the absolute difference from PI.


gruselhaus:
Lovely ❤️

PR #1247, opened 2019-04-03
Rubik's Cube code refactor. (Colors enum is added to the code)

harpreets789:


shiffman:
Oh this is awesome! I have to think about how I might like to add this to the code. I try to keep the source the same as what is in the videos, but I could also make a short video with refactoring to use `Enum`. For now would you want to submit a community contribution link so people can find your refactored code from the challenge page? 

https://thecodingtrain.com/CodingChallenges/142.1-rubiks-cube.html

harpreets789:
> ould you want to submit a community contribution link so people can fin

ya sure, I will check all the steps first, how can I make community contribution on changeling page and will do that in few days.

PR #1278, opened 2019-04-16
CC_142 [Rubik's Cube] Refactoring the 'Move' logic into an enum.

andi-huber:
- no more need to encode move sequences with characters
- new feature: random moves are such, that sequential moves do not cancel each other out

Other than that, does not add anything new, its simply a cleanup/refactoring using the power of enums. Hope it has some educational value.

Cheers!

### DIFF
--- a/CodingChallenges/CC_142_Rubiks_Cube_2/Processing/CC_142_Rubiks_Cube_2/Axis.pde
+++ b/CodingChallenges/CC_142_Rubiks_Cube_2/Processing/CC_142_Rubiks_Cube_2/Axis.pde
@@ -1,0 +1,56 @@
+enum Axis {
+    X {
+
+        void turn(Cubie[] qbs, int selector, int dir) {
+            for (int i = 0; i < qbs.length; i++) {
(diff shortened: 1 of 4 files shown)


shiffman:
These are terrific suggestions, thank you for taking the time to do this work! I like to keep the code in this repo matching the video so that viewers can follow along. There are a couple ways we could move forward to incorporate your ideas. 

1. You could submit your own version as a [community contribution](https://github.com/CodingTrain/website/wiki/Community-Contributions-Guide).
2. I could create a community "Rubik's Cube" project separate from the website that viewers could develop. I like doing this, but it often becomes difficult to manage.

What do you think?

(I'll also note that I am releasing part 3 tomorrow).

andi-huber:
Thanks Daniel!

I suspected, that this PR would not get merged, however I could not think of a more simple way to reach out to you, with a refactoring suggestion. I'm impressed, that this got feedback at all. You and your team are really doing a great job!

So I guess we can just close this PR. 
(No worries on my side.)

Keep up the good work!

PR #1282, opened 2019-04-17
Code refactored and Moves implemented

vladig98:
I refactored the code so that all rotations are in one function. I've implemented the real moves of the Rubik's cube. Colors of the cube updated.

PR #1292, opened 2019-04-21
cc-142.3 Rubik Cube

jorgef1962:
A proposed change to the cubie class.
- Paints internal faces in black
- Excludes never visible internal faces
- Excludes the central (0, 0, 0) cubie.

As an old timer, even today, with tons of MIPS and RAM measured in GB, I still have a weird feeling about wasting either RAM or processing power, so....

Regards

EDIT ADD: The face colors have been reorganized acording to the http://iamthecu.be/ site, considering the vertical reversion of the Y axis (positive down).


shiffman:
Like with #1282 I think I'd like these improvements to be linked as community contributions so that the code matches the video but viewers can easily find improvements?

shiffman:
Same here, would like to leave open for a little while, thank you @jorgef1962 for this wonderful suggestion!

PR #1300, opened 2019-04-27
minor edits to CC 144 code headers

shiffman:
@gruselhaus do you think it makes sense to include this version too here on GitHub or the links from the code to the web editor is enough?

https://editor.p5js.org/codingtrain/sketches/4DvaeH0Ur


gruselhaus:
I think it does 😎

PR #1305, opened 2019-04-30
darken some colors to meet WCAG AA

phucduong86:
Nav Before:
image
Nav After:
image

Button before and after:
image
image

image
image

Before:
image

After:
image

Before:
image
After:
image

### DIFF
--- a/assets/css/1-tools/_vars.sass
+++ b/assets/css/1-tools/_vars.sass
@@ -1,6 +1,6 @@
 // Basic Colors
 $text-color: #222
-$secondary-color: lighten($text-color, 40%)
+$secondary-color: lighten($text-color, 25%)
 
 $box-shadow-primary: #eee
 $box-shadow-secondary: darken($box-shadow-primary, 10%)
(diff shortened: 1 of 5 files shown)


phucduong86:
Also added focus style (replicated hover) for buttons to make it more visible.

shiffman:
This looks great to me! Design-wise I don't want to be picky but for the nav if there's a way to change the text color or pick something a bit more blu-ish that blends with the background sky that might be a little better?

phucduong86:
> This looks great to me! Design-wise I don't want to be picky but for the nav if there's a way to change the text color or pick something a bit more blu-ish that blends with the background sky that might be a little better?

Hmm changing the text color will affect mobile look. Maybe I'll merge this in and play with the color when I work on the mobile nav?
image

phucduong86:
This is a little more blue if you want:
image

phucduong86:
That looks great @nilsweb.
I'll put that on the todo list.

nilsweb:
I've just taken a look at the live page. Looks great! Only thing I'm missing is a hover style for the links in the navigation bar, like a color fade, pop, etc.. Maybe when we use the "bookmarks", they can pull down on hover? Just putting it here so we don't forget.

PR #1307, opened 2019-05-01
Update 144-black-hole-visualization.md

n6a1t4e:
My simulation includes presets of a random solar system, a massive orbital system (similar to the black hole), as well as a simulation of the Earth and Moon.


gruselhaus:
Looks great ❤️ Thank you for this contribution!

Issue #1227, opened 2019-03-25
Accessibility - 1.4.3 Contrast Minimum


phucduong86:
Most of the button elements and all of the navigation links do not meet minimum color contrast requirement.
```
<a href="/CodingChallenges">Coding Challenges</a> 
<a href="/Tutorials">Tutorials</a> 
<a href="/Streams">Streams</a> 
<a href="/Courses">Courses</a> 
<a href="https://github.com/CodingTrain/website" target="_blank">Github</a> 
<a href="https://youtube.com/thecodingtrain/?sub_confirmation=1" target="_blank" class="youtube">Subscribe on YouTube</a>
<a href="https://youtube.com/thecodingtrain/join" target="_blank" class="patreon">Become a Member</a>
<span class="video-date">19 Feb 2019</span>
<a href="https://youtube.com/thecodingtrain/" target="_blank" class="more-videos">Watch more videos</a>
...
```

phucduong86:
Resolved by #1311

Issue #1225, opened 2019-03-25
Accessibility - 1.1.1 Non-text content


phucduong86:
Homepage:
Missing alt-text
```
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 180 180"> <path fill="#f96753" d="M108.8135992 26.06720125c-26.468266 0-48.00213212 21.53066613-48.00213212 47.99733213 0 26.38653268 21.53386613 47.85426547 48.00213213 47.85426547 26.38639937 0 47.8530655-21.4677328 47.8530655-47.85426547 0-26.466666-21.46666613-47.99733213-47.85306547-47.99733213"></path> <path fill="#052a49" d="M23.333335 153.93333178V26.0666679h23.46666576v127.8666639z"></path> </svg>
```
Recommend adding <title>"alt text"</title> tag as first child element within <svg> element.

phucduong86:
Resolved by #1311

PR #1311, opened 2019-05-04
Accessibility update

phucduong86:
Added required tittle and attributes for Patreon logo svg.
Changed nav bar color and added some glow effect on hover/focus.


phucduong86:
This is related to #1225 and #1227.

PR #1314, opened 2019-05-05
Update 059-steering-text-paths.md

jwkoelewijn:
Implementation using Clojurescript and Quil (Clojure(script) library based on processing/p5).
Added some colouring of the particles based on their distance to their target location, as well as transitions between words.

Might be of interest to people trying to do this challenge in a functional manner rather than a more OO approach.

Enjoy!

nice-day


gruselhaus:
Thank you!

PR #1315, opened 2019-05-07
Add P5 version of CC 132 fluid simulation

Derek-Wds:
Hi Dan,

I have made a P5 version of the 2D fluid simulation in CC 132. Hope it helps! If you have time, I would appreciate it a lot if you could have a review on this.

Thanks,
Derek


cnie2:
Hi Derek, thanks for the port! I was wondering if you had performance issues, as I am struggling to get it to run at any decent frameRate :( any ideas for how it could be accelerated (maybe using WEBGL?)? Thanks!

PR #1332, opened 2019-05-09
Added new contribution

nem0-97:
Uses SVG, lets you add multiple line obstacles, ray sources, and have varying colors and angles to cast rays out of
Running Demo:
https://nem0-97.github.io/RayCasting/index.html


gruselhaus:
Looks great ❤️

PR #1336, opened 2019-05-10
[CC 146] Add contribution with geometrical distortions correction

SolarLiner:
This builds off the code at the end of the coding challenge to correct for geometrical distortions.


gruselhaus:
Great contribution

PR #1334, opened 2019-05-09
added example to CC 145-2d-ray-casting.md

quillaja:
Added info for my example "wolfenstein 3d style raycasting with randomly generated mazes". This was inspired by a [video](https://www.youtube.com/watch?v=eOCQfxRQ2pY) by Matt Godbolt about how Wolfenstein 3D did raycasting in the ancient days.

keyboard arrow keys move player (also pressing at top, bottom, left, right of screen on touchscreen).
space toggles 3D/2D view, S toggles some statistics and viewing rays in the 2D view.

PR #1338, opened 2019-05-10
moved example from CC145 to CC146.

quillaja:
CC 146 wasn't posted when I submitted my example to CC145, but is a much more appropriate match with CC146. See #1334 for the PR that originally added my example to C145. Sorry for the trouble.


gruselhaus:
No worries.

PR #1348, opened 2019-05-12
Adding script to check code variations

gruselhaus:


gruselhaus:
Thank you @Fraasi for the Idea❤️

PR #1380, opened 2019-05-22
Rewrote the navigation section.

phucduong86:
All related to navigation in mobile mode: 

- Nav toggle button and all the links are now keyboard accessible
- Using button instead of checkbox for better user experience with AT.
- Included jquery, updated CSS to make sure most the styling stay the same.
- Made change to link focus styling to differentiate from other elements.


phucduong86:
Working on fail tests.

meiamsome:
I'm not sure if we should be including jQuery for this one feature, but am open to thoughts on that

phucduong86:
> I'm with @meiamsome on this one. I don't think that it's worth to include the whole jQuery library just for the navigation menu. Also, the content of `mobileNav.js` should go into `functions.js` to reduce the number of network requests the browser has to make since Jekyll doesn't minify/bundle javascript right now.
> 
> Another thing I'm not quite sure is why we need keyboard support for the _mobile_ navigation? Maybe there's a use case I haven't thought of until now? 🤔

// To do:
- Moving mobileNav.js into functions.js
- Rewrite in vanilla JS and remove jQuery dependency

To answer your question about keyboard support for mobile navigation - keyboard support for ALL action elements on the webpage is one of the critical requirements to meet WCAG 2.0/2.1. There are screenreader and braille device that don't have access to a mouse, one group of user I can think of are the ones that use mobile device/tablets with bluetooth keyboard.
https://simplyaccessible.com/article/mobile-keyboard-support/

This article also explains why using button gives user much better user experience than using a checkbox input.

https://www.smashingmagazine.com/2017/11/building-accessible-menu-systems/

phucduong86:
Removed jQuery dependency however for some reason the code doesn't work unless I keep it at the end of the body. I'll investigate further but mobileNav.js is separate at the moment.
I'm also thinking of making the styling change to focus style from current
image
To this:
image

phucduong86:
> Removed jQuery dependency however for some reason the code doesn't work unless I keep it at the end of the body. I'll investigate further but mobileNav.js is separate at the moment.
> I'm also thinking of making the styling change to focus style from current
> image
> To this:
> image
> 
> @shiffman @meiamsome @nilsweb

Removed jQuery and merged mobileNav.js into functions.js
Also implemented the style change above.

PR #1399, opened 2019-06-07
Corrected mut function in Neural_Network.pde file

sushantPatrikar:
Hello! I think there is a little error in 'mut' function of 'Neural_Network.pde' file in coding challenge 100. From my knowledge the if condition on line 69 should be:

`if(random(1) < rate)`
instead of:
`if(random(1) > rate)`

Suppose the rate is 0.9, then what our intention is to mutate 90% of the time. But according to me, the function is performing other way round right now i.e the probability that the random(1) > 0.9 is just 10%.
Also when I checked the P5 version of this coding challenge, I found the 'mutate' function correct there.
Here are the snippets from both the versions.
This snippet is from P5 version which I think is correct:

image
 

And this is from Processing version:

image
 

Correct me if I'm missing something here.
Cheers!


gruselhaus:
@GypsyDangerous What are your thoughts?

GypsyDangerous:
I agree it needs to be changed.

PR #1449, opened 2019-07-04
Update 145-2d-ray-casting.md

Damo676767:
Small change to the code to better reflect how the intersection of a line and a line segment is being done

https://github.com/Damo676767/Random-Pieces-of-Code/blob/master/ray.js

### DIFF
--- a/_CodingChallenges/145-2d-ray-casting.md
+++ b/_CodingChallenges/145-2d-ray-casting.md
@@ -92,6 +92,12 @@ contributions:
       url: "https://simontiger.com/"
     url: https://editor.p5js.org/simontiger/present/F6TCHAZs_
     source: https://editor.p5js.org/simontiger/sketches/F6TCHAZs_
+    - title: "Minor optimisation ray.js to better suit math being done"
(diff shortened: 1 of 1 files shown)


shiffman:
Thank you for this! @Damo676767 would you mind elaborating a little either in a comment here or commenting in your source on what exactly you changed / updated?

Damo676767:
Hi, thanks for replying

Certainly. In your function where you calculated the distance between a ray
and a boundary, you calculated a 2nd point on the ray to plug into the
Wikipedia formula. That formula then uses that point to calculate the
gradient (sort of) of the ray.

As you essentially used the gradient of the array to calculate that point.
I modified the formula to use the already stored gradient rather than
recalculating it.

I really enjoy your content and at this exact moment am sitting down going
through your tutorials on p5js. I have done a reasonable amount of coding
with c, java, MATLAB, and a few other languages but am really enjoying
working out p5js. So a big Thank You to you for that.

Thanks
Damo

shiffman:
Thank you!!!

PR #1452, opened 2019-07-06
comments are added

Rashmi-278:


gruselhaus:
Thanks!
Do you mind adding these at the p5 version as well?

Rashmi-278:
> Thanks!
> Do you mind adding these at the p5 version as well?

Sure I will do it :)

PR #1460, opened 2019-07-11
Coding Challenge 148

shiffman:
Hoping to release this today after finalizing thumbnail and captions. This pull requests includes the markdown file and the p5.js code.


shiffman:
With the new members only setting on YouTube, I can merge these before the video is 100% finalized and anyone who finds the video will be encouraged to sign up. Maybe that's annoying, I'm not sure!

(Further explanation: I am using 3PlayMedia for captions and it can take 4-5 days for those to be ready. So I do members only while the captions are being prepared then public when ready.)

PR #1461, opened 2019-07-12
Adding Processing Version of CC148

gruselhaus:


CodingChallenges/CC_148_Gift_Wrapping/Processing/CC_148_Gift_Wrapping/CC_148_Gift_Wrapping.pde
```diff
@@ -0,0 +1,89 @@
...
+
+void setup() {
+  size(500, 500);
+  int buffer = 20;
+
+  for (int i = 0; i < 50; i++) {
+    points.add(new PVector(random(buffer, width - buffer), random(buffer, height - buffer)));
+  }
+
+  Collections.sort(points, VEC_CMP);
```
shiffman: Wow! This is some fancy Java! I'm happy to keep this, just noting that beginners coming to this might be confused so we could alternatively just put in a simple selection sort algorithm or even simpler still just an algorithm to find the left most point. What do you think @gruselhaus?

gruselhaus: Added.

Issue #1480, opened 2019-07-24
New Contributions Tutorial Video


shiffman:
This is currently just unlisted on the channel (working on the thumbnail, etc.) but a link to it should be added to all the places where the site references adding your version of a challenge, etc.!

https://youtu.be/fkIr0inoQ7Y

(I went a little crazy with the labels for this issue, 😹)

PR #1523, opened 2019-07-27
Moving the Wiki to own Markdown files

gruselhaus:


jonnytest1:
https://github.com/CodingTrain/website/blob/ae54cb7a51a9be4c10aa30cd9928c0a043cec5b5/Guides/Website_Contribution.md :
id dttn think you need the following steps: 
`bundle exec jekyll build`
`python -m SimpleHTTPServer`

nilsweb:
> id dttn think you need the following steps:
> `bundle exec jekyll build`
> `python -m SimpleHTTPServer`

In fact, you don't _really_ need them but I tried to explain a little bit how Jekyll works and what the process behind building a website with Jekyll is. That's why right below those two commands I've put the 'development-mode' command which watches your changes and updates the server automatically.

nilsweb:
@gruselhaus 

Nice work so far! :tada:
I'll quickly go ahead and add the headlines to the pages as well since they are still missing. ;)

I would have completely forgotten to exclude the guides from the Jekyll build but it made me think that it might be nice to render the guides on the website as well so that when you click on the 'see how' link for adding a community contribution, it will link to the version on the website (btw, the link needs to be updated as well, we should't forget that).

nilsweb:
@gruselhaus Please review! 😋

List of changes:
- The website now renders all guides to HTML
- The HTML title tag got updated and now shows the correct title for each page for better SEO and accessibility
- We can now develop locally with Docker to circumvent scary errors due to Ruby's native extension nonsense

gruselhaus:
Yup. Looks good to me. Let's merge! 🚀

gruselhaus:
#1480

PR #1563, opened 2019-07-29
Including 10PRINT with a maze crawler by Volfegan

volfegan:
Including 10PRINT with a maze crawler function by Volfegan


gruselhaus:
Duplicate of #1564

PR #1564, opened 2019-07-29
Added 10PRINT with a maze crawler function

volfegan:
Including 10PRINT with a maze crawler function by Volfegan


gruselhaus:
This is awesome!

PR #1573, opened 2019-07-30
TSP with 2-opt

BIGfoot496:
A vizualization of 2-opt algorithm solving the travelling salesman problem.

### DIFF
--- a/_CodingChallenges/035.1-tsp.md
+++ b/_CodingChallenges/035.1-tsp.md
@@ -48,6 +48,10 @@ contributions:
       url: https://github.com/storealutes
     url: http://chaseacton.com/sterling/tsm/
     source: https://github.com/storealutes/travelingSalesMen
+  - title: "TSP with 2-opt"
+    author: "Elijah Beregovsky"
+    url: https://editor.p5js.org/BIGfoot/full/EqDo-eigW
+    source: https://editor.p5js.org/BIGfoot/sketches/EqDo-eigW
 ---
 
 In Part 1 of this multi-part coding challenge, I introduce  the classic computer science Traveling Salesperson problem ("traveling salesman" (sic) for search) and discussing the pitfalls with a brute force solution.


shiffman:
Wow! I'm not familiar with this algorithm. Any recommendations for resources? Worth a separate coding challenge?

BIGfoot496:
Hello!
The basic layout of algorithm is this:
-Start with a random path
- Cycle through all pairs of disjoint edges of a path
- For each pair there can be two other disjoint pairs of edges, but only one of them does not split the path in two. If this other pair’s total length is less, than that of the current pair, change current pair of edges to the better one and repeat the previous step.
- When the previous step doesn’t change the path, we’re done.

I came up with this algorithm myself, but later found out that I had invented the wheel. Here’s the Wikipedia page: https://en.m.wikipedia.org/wiki/2-opt

I think that 2-opt is really simple, and what’s cool about it, is that the paths it produces resemble highly of what a human may draw, if he was to solve the tsp manually. They are not by any means ideal, but they are good enough for you to actually consider using them in real-life tsp. That alone indicates this deserves a quick Coding Challenge. Would be happy to see it.

And 2-opt is great to build upon. It can clean up the obvious blunders of a greedy algorithm like Dijkstra. Or you can use swapping pairs of edges as a decent mutation for a genetic algorithm, which is nearly not as random as the one you used in your Coding Challenge. But the best thing, in my opinion, to do with it, and I’d really really like to see a Coding Challenge on THIS algorithm, is using it as a mutation in a simulated annealing. Here’s it: https://en.m.wikipedia.org/wiki/Simulated_annealing

Big thanks for your channel! Visualizing algorithms is perfect for understanding them, and your videos are always a splendid point to start tinkering with new algorithms and to build upon!

BIGfoot496:
The basic k-opt algorithm and improvements of it are well described here: http://akira.ruc.dk/~keld/research/LKH/LKH-2.0/DOC/LKH_REPORT.pdf

PR #1574, opened 2019-07-30
Add community contribution

omareq:
Animation of multiple convex hull generation algorithms.

https://omareq.github.io/convex-hull/


omareq:
Thanka @gruselhaus for fixing that indent

Issue #1587, opened 2019-07-31
CC_098.3_QuadTree_Collisions not working properly if canvas size is not square


volfegan:
Quadtree for processing only works properly if the canvas is a square size (for processing). If the width and height are different, the sectors start to not be aligned and miss the created points in the gaps.
The gaps between the quadtree sectors are proportional to the difference between width and height.
Image for bug occurring:
https://twitter.com/VolfeganGeist/status/1156633440252506112

Code to visualize the bug:
https://gist.github.com/volfegan/f1f3676dfc0e6ccc2c4593e7d74691dd

The same issue occurs on CC_098.1_QuadTree for processing

PR #1589, opened 2019-08-01
Fixing issue on CC_098.1_QuadTree not working properly if ...

volfegan:
…canvas size is not square
This merge is to fix CC_098.1_QuadTree

Another merge will fix CC_098.3_QuadTree_Collisions
Class QuadTree was passing (width, height) to the Rectangle constructor and the Rectangle was using it inversed as (height, width).
Basically, QuadTree was asking for ▭ and the Rectangle was doing ▯.

This will fix the issue: #1587


gruselhaus:
Closes #1587

PR #1591, opened 2019-08-01
redoing Fix issue on CC_098.3_QuadTree_Collisions

volfegan:
Sorry I used the same title and was marked as duplicated. Copy-paste laziness from a long day.
This merge is to fix CC_098.3_QuadTree_Collisions
The other merge is to fix CC_098.1_QuadTree

QuadTree not working if canvas size is not square
Class QuadTree was passing (width, height) to the Rectangle constructor and the Rectangle was using it inversed as (height, width).
Basically, QuadTree was asking for ▭ and the Rectangle was doing ▯.
This will fix the issue: #1587


gruselhaus:
@shiffman may we merge?

PR #1592, opened 2019-08-01
TSP with simulated annealing

BIGfoot496:
It's a version of SA used for solving TSP using a 2-opt move (shifting pairs of edges) as a mutaion. Constants are somewhat random, so it probably can be done much better, but the results look decent.


BIGfoot496:
It performs slightly better than "greedy" 2-opt, but running time is ten times longer, so, in my opinion, this improvement is negligible, but again, the constants are pretty random and can be tweaked into something more usable.

PR #1605, opened 2019-08-05
Added `js-markov`

Edwin-Pratt:
`js-markov` was my first ever JavaScript library which I have made. I got inspired by this challenge so much that I have improved the code and even started teaching other 8th graders about Markov Chains!


shiffman:
Wonderful work!!

PR #1612, opened 2019-08-10
Added my contribution to the list

Guy-Yome:
This was improvised and I did not do any research so I can try to come up in my own way with a solution. You can sadly not see it step-by-step since I realized I could do it right at the end. Oops! My bad ;)


Guy-Yome:
Also, click to add new points

PR #1614, opened 2019-08-11
TSP with Genetic algorithm

BIGfoot496:
GA for TSP with a 2-opt move as a mutation and no crossover. As a fitness function I used f(s) = av - len(s), where len is length of the path s, and av is a value more than most of the pathlengths and of the same order. After some constant tweaking I managed to make it work as good as 2-opt and faster than simulated annealing.


BIGfoot496:
On the larger input sizes the constants must be tweaked again, because the speed of the algorithm falls drastically.

BIGfoot496:
av gets smaller as the algorithm progesses

BIGfoot496:
By the way, does anybody know, how to implement speciation in a GA for travelling salesman? It could probably improve performance.

BIGfoot496:
Most of the paths in the population of my algorithm are alike and "biodiversity" may help the algorithm produce better results. As for now, it's just another version of a gradient descent.

PR #1708, opened 2019-09-07
added 3 or more player support to CC_149

rogerwim:
and a bit of cleaning up


rogerwim:
i have it almost perfect

PR #1725, opened 2019-09-15
.htaccess file added

realabbas:
# .htaccess file
For better browser leverage caching 

Comparison report between current website without browser caching and after adding .htaccess file 

image

**Comapre Report:** https://gtmetrix.com/compare/2nrQJYW0/HzYNHkAF

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -1,0 +1,19 @@
+<IfModule mod_expires.c>
+  ExpiresActive On
+
+  # Images
+  ExpiresByType image/jpeg "access plus 1 year"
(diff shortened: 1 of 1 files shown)


gruselhaus:
Does GitHub Pages support htaccess? I don't think so...

gruselhaus:
Does the general performance depends on the hosting provider? 🧐

meiamsome:
@gruselhaus no, I don't think it does - this file would do nothing for thecodingtrain.com - perhaps it would for anyone running it via Apache, but I'd suggest that this file shouldn't be in the repo

PR #1737, opened 2019-09-30
Add sort.khoi.io

khoi:
I've recently found out about Coding Train videos and loving it <3. This contribution adds my visualizations on the sorting challenges.


shiffman:
These are fantastic, thank you for sharing! Are all the rows sorted individually or is there a relationship between them? I like the idea of sorting hues, would you mind if I did a future coding challenge inspired by your examples?

khoi:
Each row contains the same values set and are shuffled and sorted individually. 

> would you mind if I did a future coding challenge inspired by your examples?

not at all, I'm looking forward to seeing it

PR #1740, opened 2019-10-01
changed style for dropdown links

GypsyDangerous:
I don't know if this will break other things but i think looks better with a gap between the items.


shiffman:
Could you add a comment here with some screenshots to show the difference? Thank you!

GypsyDangerous:
before
after

GypsyDangerous:
First is before, second is after

vimkat:
Thanks for this, @GypsyDangerous! 🎉 
I'd add some form of background so the buttons don't look that "lost" floating around but thats purely a design choice. I'm happy for other opinions!

GypsyDangerous:
@vimkat I agree that a background would look better but im not sure how to do it.

vimkat:
> @vimkat I agree that a background would look better but im not sure how to do it.

You could try giving the `ul` some `background` and maybe padding.

GypsyDangerous:
another picture with a little higher quality

GypsyDangerous:
and here is one with window full sized

GypsyDangerous:
i made the background more purpleish/pinkish to fit the style of the website

GypsyDangerous:
oops

GypsyDangerous:
new background
ezgif com-video-to-gif (1)

assets/css/3-layouts/_video.sass
```diff
@@ -94,7 +94,7 @@
             border-radius: 0 0 3px 3px
             overflow: hidden
             transition: all 100ms
-            background: rgba(145, 141, 141, 0.65)
+            background: #a557f147
```
phucduong86: This is not a valid Hex value.
Please also check color contrast to make sure it's above 4.5:1 radio for white text.

GypsyDangerous: i used [this](https://contrast-ratio.com/#white-on-rgba%28162%2C%2083%2C%20200%2C%200.37%29) site to check the ratio and got it to a ratio the site said was good.

GypsyDangerous:
here is another idea for background color, make it the same color as the button.

Annotation 2019-10-02 214355

vimkat:
I _really_ like the second version! 🎉

GypsyDangerous:
@vimkat the one where it is the same color as the button?

GypsyDangerous:
@shiffman @phucduong86

shiffman:
I like all of these! As long as they meet the WCAG accessibility guidelines I am good to merge!

phucduong86:
Hi, 
There are a few issues with the dropdown buttons in regard to accessibility:
- It can only be activated by hovering, so user using keyboard are not able to access the links within the drop down.
- The color contrast in the drop down does not meet minimum requirements for WCAG 2.0.

GypsyDangerous:
in regards to the first point, how could we add keyboard accessibility to it? and the second point, what colors should i be checking contrast on, because the background color `#9659A7` has a 4.9:1 contrast with the text color `#FFFFFF`

GypsyDangerous:
I just modified the background.  now it looks like this and according to [this](https://webaim.org/resources/contrastchecker/) site the contrast should be good

GypsyDangerous:
@phucduong86 what part of this should I change to make it meet the contrast requirements?
66093301-c61cb480-e55d-11e9-864b-113416f4a356

phucduong86:
> @phucduong86 what part of this should I change to make it meet the contrast requirements?
> 66093301-c61cb480-e55d-11e9-864b-113416f4a356

Hi @GypsyDangerous, 
The background in the inside button

image

The element that are only shown when hover/actived are sometimes missed by automated checker.
When I inspected the element I got the color #AB7AB9 that is only 3.37:1 with white text:
image

image

For the question about where to check color contrast - most contrast checkers are quite accurate if we give them the correct color value. ColorZilla chrome extension does a good job picking colors usually. Sometimes the colors are altered with filter/opacity/overlay and we kind of have to make the judgement call if it is accurate or not. If it doesn't seem like it has the right value I often take a screenshot and sample the actual color then send them through the contrast checker. 

On the keyboard accessibility - you need to make sure (programmatically in Javascript) that:
- The main button is focusable
- The main button listen to keyboard event
- The main button activate/deactivate the expanded state (visually) on keyboard action and also announcing this state to end user (users who are blind for example) via aria properties.

I've implemented similar function for the mobile nav menu if you want a reference.
This is also a very valuable resource on the specific topic:
https://www.w3.org/WAI/tutorials/menus/flyout/

Cheers!

GypsyDangerous:
okay I fixed the color issue. I have a solution for the keyboard issue but i dont know how to add to the repo because i can't find where to change the html

GypsyDangerous:
never mind i found it and made the changes in #1759

GypsyDangerous:
I'll add pictures this afternoon when i get home.

gruselhaus:
@GypsyDangerous what is the status of this PR?

GypsyDangerous:
@gruselhaus I have been working on this on and off lately, I've studied the WCAG standards to try and get a better understanding and I'm trying to find a good color scheme that meets the standards. It has been really difficult to work since I am currently unable to build the website from my fork but it will get a lot easier when I upgrade my computer and get Linux in early January. As for the issue with the buttons not being accessible as mentioned in #1759 I am still stuck on finding a good solution for the problem

PR #1744, opened 2019-10-01
darken the navbar on hover (with a smooth transition)

AnTheMaker:
this makes reading (and clicking) on a link in the navigation bar a lot easier (because it's better readable) on desktops.


GypsyDangerous:
Can you post before and after pictures of it?

AnTheMaker:
@GypsyDangerous here is a comparison:
Before: 
codingtrain_before

After:
codingtrain_video_gif

vimkat:
Thanks for helping out, @AnTheMaker!
Also tagging @phucduong86 to have a look at accessibility.

assets/css/2-base/_navigation.sass
```diff
@@ -46,13 +46,20 @@
...
+    &:hover
+      background: rgba(9,20,154,0.74)
 
     @media (max-width: $mobile)
       width: 0%
       height: auto
       top: 10rem
       position: fixed
       background: none
+
```
phucduong86: Changes look great!
I'm just not sure if this is necessary - quick test on the browser looks like it doesn't really make a difference. Otherwise definitely an improvement :)

PR #1746, opened 2019-10-02
Add a text for showing best record distance in TSP Challenge.

amirfeqhi:


amirfeqhi:
Thank you so much @shiffman ❤️

PR #1756, opened 2019-10-06
Ukulele Tuner contribution: Bass/Guitar Tuner with VU style

tonetlds:
This is my little experiment. I have added the p5.Amplitude() too.


shiffman:
Nice work!!

PR #1767, opened 2019-10-09
Update 126-toothpicks.md

tht-jxny:
This contribution implements a infinite zoom out to render more and more toothpicks. It eventually takes a while to compute ^^.


tht-jxny:
https://editor.p5js.org/tht-jxny/full/HJXcWqoW4 is the link

PR #1779, opened 2019-10-13
removing references to siraj's material due to scandals

GypsyDangerous:


shiffman:
Thank you @GypsyDangerous. 😞

PR #1805, opened 2019-10-20
Update link-list.html

rishraj2000:
was just doing test pull, will contribute at the earliest


violetcraze:
@rishraj2000 It's great to see that your getting familiar with making pull requests. Is this fixing a problem or did you do it to just so you could make a pull request? Both are fine just like to know if it should be added.

Issue #1834, opened 2019-10-26
Navigation bar resizing


Tekman03:
Since new buttons have been added to the navigation bar in recent weeks, the bar now over extends on devices in between 750ish px and 550px.

Was wondering if simply putting a new sass var for the breakpoint of the nav at 750px? Maybe setting it to $tablet-big(: 920px) would look the best? Or does somebody want an intermediary step between desktop and mobile?

I am happy to do this if it sounds fine?

GypsyDangerous:
I'm gonna close this because that merge seems to have fixed it, feel free to reopen if needed.

PR #1835, opened 2019-10-26
Update breakpoint for nav bar

Tekman03:
As mentioned in Issue #1834 

$mobile breakpoint was not sufficient to stop the nav bar overflowing. Have updated to $tablet-big(:920px) for sufficient space (as of current nav layout).


violetcraze:
Thanks for catching this bug. Been thinking about redesigning navbar but this is a great fix for the time being.

PR #1839, opened 2019-10-27
Create LICENSE

shiffman:
Somehow we're missing a license on this repo! Is MIT ok with all?


gruselhaus:
@meiamsome What do you think?

PR #1851, opened 2019-10-28
Adding latest cabana video to landing page

mikakruschel:
Hey! I'm not sure if it is intended that the latest cabana video does not show up on the landing page. But I changed it in this PR! I'm new to Jekyll so there might be better ways to implement it.

codingInTheCabanaOnLandingPage

### DIFF
--- a/_jekyll/includes/1-tools/sort-videos.html
+++ b/_jekyll/includes/1-tools/sort-videos.html
@@ -3,6 +3,7 @@
 
   Parameters:
     videos Video[]              Array of videos that will be sorted.
+    sortByDate bool             Wether to sort the videos by date (default: false).
     reverse bool                Wether to reverse the order in which videos are storedin the array (default: false).
     future bool                 Wether to include future videos (default: false).
     [1.1] seriesPath string     The series path by which the videos will be filtered.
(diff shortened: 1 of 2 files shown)


gruselhaus:
I've created a simple Badge for the cabana vids so viewers can differentiate between CC-Videos and Cabana-Videos on the Startscreen.

What are your thoughts? @shiffman @violetcraze @mikakruschel 

Bildschirmfoto 2019-10-29 um 15 40 02
Bildschirmfoto 2019-10-29 um 15 40 05

shiffman:
Hi @gruselhaus, nice work!! Maybe it's not necessary for the cabana page itself but more for the homepage to distinguish? I like it but my concern is that it obscures the thumbnail image quite a bit? I am working with Jason Heglund on some new design sketches for the Cabana series so there might be something there too!

gruselhaus:
> Hi @gruselhaus, nice work!! Maybe it's not necessary for the cabana page itself but more for the homepage to distinguish? I like it but my concern is that it obscures the thumbnail image quite a bit? I am working with Jason Heglund on some new design sketches for the Cabana series so there might be something there too!

@shiffman ok, thank your for your feedback. Nevertheless let's merge this PR and pick up the "Cabana" Banner later on.

PR #1870, opened 2019-10-30
Adding FAQ Section

violetcraze:
Been talking with @cdslls about adding a FAQ section to the website.

This is a jumping off point. I haven't added much content or any sorting features besides requiring a section for each questions. The sections are displayed in the order they are added to the list of sections. The questions are displayed within each section in the order they are added to the faq list. Hope that makes sense.

I think the unit testing for this page could be improved the error messages it produces aren't very decipherable. Not super familiar with the library that is being used so I didn't look into it right now. @meiamsome and @gruselhaus Any advice on how it could be improved would be great.

Basically the format is what's currently in `_FAQ/index.md`. Sections has a list of titles that are a string. `faq` is a a list of questions that require a question, answer, and section with optional links. The sections are case sensitive.

Didn't include in navigation because it doesn't have content and the navbar needs to be restructured as it is.

Here's a screen shot of what the page would look like.

Screenshot of FAQ page


shiffman:
This is great and a perfect start! Design-wise the centered headings don't fit well maybe? Maybe it's the centering or maybe it's that the font size is so close to the questions. But I am not the best visually here and just happy that the information will be out there! For reference, @3b1b has a nice FAQ:

https://www.3blue1brown.com/faq

GypsyDangerous:
@violetcraze is this up to date with the new dropdown menus you added for the headbar

GypsyDangerous:
also, I agree with @shiffman that the headers would look better not centered

gruselhaus:
@violetcraze any updates?

violetcraze:
@gruselhaus Haven't updated this recently. I'll close it and open a new pr when I get around to working an faq more.

PR #1878, opened 2019-11-02
Fix release dates

mikakruschel:
I wrote a script ([link to gist](https://gist.github.com/mikakruschel/737e33f1c620c8a9469258449b68d3aa), that I could include it in this PR if you want) that fixes the release date for every video /`.md` files using the Youtube Data API.


violetcraze:
@mikakruschel This is great! Not sure if we should include the script because it could cause some problems if it was running silently and changing things. But it would make a great unit test if instead of automatically changing values it would just throw an error if the dates don't match.

PR #1881, opened 2019-11-04
rearranged footer links

GypsyDangerous:
Since the new `back to top` has the same background color as the `shiffman.net` footer link I moved it from the far right to the far left, I don't if this is the best solution but it is the easiest.

Before Before

After After

Full footer after full after

### DIFF
--- a/_jekyll/includes/2-base/footer.html
+++ b/_jekyll/includes/2-base/footer.html
@@ -4,8 +4,8 @@ This file will be included after every page.
 
 <footer>
   <div class="links">
+    <a href="http://shiffman.net/" class="shiffman-net"><span>shiffman.net</span></a>
     <a href="http://natureofcode.com" class="nature-of-code"><span>Nature of Code</span></a>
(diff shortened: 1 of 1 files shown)


GypsyDangerous:
Sorry for the poor image quality I can't take screenshots on my computer at work.

Fraasi:
whoops. I noticed this when I first made this for the new site and meant to fix it with more distinct border or drop-shadow, but forgot the whole thing.
I guess this is as good solution as any.

GypsyDangerous:
@Fraasi am I good to merge or should I wait for @vimkat to look at it?

Fraasi:
I don't think the order of the footer links matter. So I think this is good to go, unless you want to wait for some design-wise comment from someone else.
Anyway what I like about this site is that we can change this again next week if this somehow doesn't work for everybody 😃

vimkat:
Looks perfect, @GypsyDangerous! 👍

PR #1882, opened 2019-11-04
changed patron to member

GypsyDangerous:
see #1879

PR #1883, opened 2019-11-05
Updating support section from patron to member

GypsyDangerous:
this is the follow up to #1882 

here is a picture of the end result
Annotation 2019-11-05 172647

I am unable to build the site from the repo on my current home computer so I can't guarantee that it works but I can make any necessary changes.


vimkat:
Awesome! Thanks a lot, @GypsyDangerous!

PR #1889, opened 2019-11-07
Horizontal version of TM header image

shiffman:
I think the page might look better with the TM header taking up less space. My local jekyll install seems to be broken so I couldn't actually test this. . . throwing caution to the wind!


shiffman:
Going to try merging this, my local jekyll is so broken and this will actually be a faster way to test, hah!

gruselhaus:
@shiffman just FYI: you can build the website with a prebuild docker image using docker-compose => `docker-compose up`

shiffman:
oh, amazing, thank you!!!

PR #1895, opened 2019-11-07
Testing gh-action env vars console.log

anuraghazra:
just testing.


anuraghazra:
Opps, my bad.. sorry.

not it won't because it will only trigger on `push` events 
closing :rofl: 

Sorry. but i'm still curious if it can be logged when pushed

PR #1906, opened 2019-11-09
Add highlight for links to guest tutorials and Teachable Machine videos

mikakruschel:
I think it is easy to overlook the links to the guest tutorials and the Teachable Machine videos, so I highlighted "guests" and "Teachable Machine":

image

### DIFF
--- a/_Tutorials/index.md
+++ b/_Tutorials/index.md
@@ -5,6 +5,6 @@ layout: series-index
 
 Here you can find all tutorials made by Daniel Shiffman on TheCodingTrain.
 
-If you are searching for tutorials made by guests, you can check them out [here]({{ site.baseurl }}{% link _GuestTutorials/index.md %}).
+If you are searching for tutorials made by <highlight>guests</highlight>, you can check them out [here]({{ site.baseurl }}{% link _GuestTutorials/index.md %}).
 
-If you are looking for the Teachable Machine videos they are available [here]({{ site.baseurl }}{% link _TeachableMachine/index.md %}).
+If you are looking for the <highlight>Teachable Machine</highlight> videos they are available [here]({{ site.baseurl }}{% link _TeachableMachine/index.md %}).
--- a/assets/css/2-base/_typeface.sass
+++ b/assets/css/2-base/_typeface.sass
@@ -38,5 +38,5 @@ h3
   font-size: 1.8rem
 
 highlight
-  color: $ct-green
+  color: $ct-blue
   font-weight: bold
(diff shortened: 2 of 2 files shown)


_Tutorials/index.md
```diff
@@ -5,6 +5,6 @@ layout: series-index
 
 Here you can find all tutorials made by Daniel Shiffman on TheCodingTrain.
 
-If you are searching for tutorials made by guests, you can check them out [here]({{ site.baseurl }}{% link _GuestTutorials/index.md %}).
+If you are searching for tutorials made by <highlight>guests</highlight>, you can check them out [here]({{ site.baseurl }}{% link _GuestTutorials/index.md %}).
```
shiffman: For accessibility, we should not use generic terms like "here". People who browse the internet with a screen reader read lists of the links and so meaningful text should always be included. 

For more see: https://www.washington.edu/accessibility/links/ and https://www.w3.org/TR/WCAG20/#navigation-mechanisms-refs

shiffman:
Ooops, I was too slow, if we could correct in a separate pull request that would be great!

_Tutorials/index.md
mikakruschel: The actual links are still called "here", it's only the texts "guests" and "teachable machine" which are highlighted

mikakruschel:
Sorry, my PR title isn't accurate, only some text is highlighted not the actual link.

_Tutorials/index.md
shiffman: Apologies for confusion it's the use of "here" that is a problem. The accessible way to create a link is as follows:

```
If you are searching for tutorials made by <highlight>guests</highlight>, you can check them
out on [the guest tutorial page]({{ site.baseurl }}{% link _GuestTutorials/index.md %})
```

There's probably a shorter way to phrase the above to reduce redundancy too!

mikakruschel: How about:
```suggestion
For even more tutorials, you can visit the [guest tutorials page]({{ site.baseurl }}{% link _GuestTutorials/index.md %}).
```
And
```suggestion
If you are interested in creating your own machine learning models, have a look at the [Teachable Machine page]({{ site.baseurl }}{% link _TeachableMachine/index.md %}).
```
English is not my native language so I'm open to any critic.

shiffman: @violetcraze would you mind taking a look at the above discussion and adjusting the language on the site? Thank you! ❤️

violetcraze: Looks great to me!

PR #1918, opened 2019-11-13
Update sketch.js

jobintweets:
Changed var to let

### Community Contribution

### Sharing

**Instagram:**

**Twitter:**


jobintweets:
> While generally speaking I prefer the code to match exactly as it is in the tutorial, the use of `var` is so out of date that updating the examples makes sense to me!

Really happy to contribute in any way possible. There are quite a few places where var is used. It would be rather nice to update those ones as well.

PR #1929, opened 2019-11-16
configuring eslint for ci testing

violetcraze:
Opening up this pull request to discuss adding eslint to ci tests keep track of example code.

Got the current config file running on coding challenges and cabana. Disabled a fair number of the checks that would require large diffs. Seems like those might be better tackled one in separate pull requests but left many of the rules it would be good to add in comments.

Planning to add Courses and Tutorials over the weekend before setting up with github actions and merging.

Any suggestions are welcome.


CodingChallenges/CC_089_langtonsant/P5/sketch.js
```diff
@@ -1,87 +1,87 @@
...
+let dir;
+
+let ANTUP = 0;
+let ANTRIGHT = 1;
+let ANTDOWN = 2;
+let ANTLEFT = 3;
+
+function setup() {
+  createCanvas(400, 400);
+  grid = make2DArray(width,height);
```
shiffman: My autoformatter will add a space here like:

```suggestion
  grid = make2DArray(width, height);
```

CodingChallenges/CC_089_langtonsant/P5/sketch.js
```diff
@@ -1,87 +1,87 @@
...
+  x = width/2;
```
shiffman:
```suggestion
  x = width / 2;
```

CodingChallenges/CC_089_langtonsant/P5/sketch.js
```diff
@@ -1,87 +1,87 @@
...
+    y--;
+  } else if (dir == ANTRIGHT) {
+    x++;
+  } else if (dir == ANTDOWN) {
+    y++;
+  } else if (dir == ANTLEFT) {
+    x--;
+  }
+
+  if (x > width-1) {
```
shiffman:
```suggestion
  if (x > width - 1) {
```

violetcraze:
Yep as far as I know linting is mainly used for finding mistakes that might cause bugs, code style, and code formatting. The way it's setup up right now it only really does the first thing. But planning to enable prettier, for the formatting, in a separate pull request which should catch all of the things. Don't think we should add much in the way of rules that enforce style because it doesn't make sense for this project.

PR #1932, opened 2019-11-17
Prettier Added to Lint and Fix on Existing Files

violetcraze:
This is a rather large change to existing files. Could look into another setup which might offer more customization as the current package for prettier doesn't offer really any options from the default settings. Looks like the more recent code has 2 space indentation which matches prettier but some of the older stuff was done with 4 space indentation. The other thing that I found unusual is the semi colons at the end of nested functions. @shiffman let me know what you think. I'm happy to look int other options for formatting.


gruselhaus:
@shiffman what are your thoughts? May we merge this PR?

gruselhaus:
@violetcraze are you able to resolve those conflicts?

violetcraze:
Opened a new pull request because it was easier than trying to rebase and then run fixes again.

PR #1959, opened 2019-12-02
Add guest page for Shirley Wu

mikakruschel:
I added the guest page for Shirley Wu and added my community contribution. 

127 0 0 1_4000_GuestTutorials_012-shirley-wu-introduction-to-d3js html

### Community Contribution

[Observable Notebook](https://observablehq.com/@mikakruschel/film-flowers-variation)
### Sharing

- [X] I would be happy to have my project shared on The Coding Train's social media!

### DIFF
--- a/_GuestTutorials/012-shirley-wu-introduction-to-d3js.md
+++ b/_GuestTutorials/012-shirley-wu-introduction-to-d3js.md
@@ -1,0 +1,37 @@
+---
+title: "Introduction to D3.js with Shirley Wu"
+video_number: 12
+date: 2019-11-15
+video_id: rQroxWLZiCo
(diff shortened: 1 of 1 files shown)


violetcraze:
Thanks for adding this page! Will have to update it when the edited video comes out but looks like everything else is great.

shiffman:
@violetcraze I don't think we are going to do an edited version of this, but rather just leave it as a live stream in all it's raw glory! There is the question of what to do about the fact that her appearance was two episodes. I suppose we could just do the usual Part 1 and Part 2 as separate pages, but something to consider for long-term is whether we want to approach multi-part videos differently?

PR #1966, opened 2019-12-05
Security Fix for activesupport

violetcraze:
Recommended upgrade by the github security feature.


violetcraze:
This didn't work tried doing a reinstall of gems as well but, that just reverted back to the old versions. Don't know it can't be updated because it's a few levels deep and the other libraries haven't patched it yet or if I'm doing this wrong. Regardless I'm closing this request because it won't fix the problem.

PR #1971, opened 2019-12-06
updating ml5 regression description

shiffman:
New description for ml5 regression video


shiffman:
Also, I can't figure out why this video doesn't show up on the associated courses page, any ideas @violetcraze or @gruselhaus? Is it because the date is 12/6?

Screen Shot 2019-12-06 at 2 38 04 PM

violetcraze:
Yeah I ran into this problem with the teachable machine videos. If the release date hasn't happened yet then sometimes it wouldn't show up. [Looks like the template](https://github.com/CodingTrain/website/blob/1f38acc7e61059931c464241f34a6f8b04917b72/_jekyll/includes/3-modules/video-card.html#L14) is only setup to handle dates in the future it they are live stream. Might be worth fixing if we continue posting pages before the official release date.

violetcraze:
You should be able to go to the direct link if you want to check it. Or you can temporarily switch the date if you're building locally.

PR #1976, opened 2019-12-09
port of blob tracking to p5js

shiffman:
This is a port to p5.js of the Blob Tracking code that goes with this video. I get asked for this a lot so thought I would port it and add it to the site. @violetcraze @gruselhaus not sure if there's a way to include on the actual webpage for this video (if there even is one!), but at minimum here it is!


violetcraze:
Looks like we're missing pages for session 9 and up on learning processing videos. I'll open an issue about adding those pages and invite others to work on it. Otherwise I can get to it later this week or next.

Even with it being on a page for a processing video I'm not sure how easy it will be to find if someone is looking in the p5 videos though so it would be good to come up with a way to access it from there.

I think it would be fine to add this though.

PR #1977, opened 2019-12-09
Adding Page for Tic Tac Toe Minimax

violetcraze:
@shiffman Check to make sure the description looks good.


shiffman:
Just making a note here that this should not be merged as the video isn't finalized yet (and the video id will change). I'll update here when ready!

shiffman:
Video is ready but set to members only, merging. 

CC154_tictactoe

PR #2011, opened 2019-12-24
Navigation Bar Drop Down Menus and Accessibility

### LINKED ISSUE #1750: It is not easy to access content and core contribution guides on the website.
It is very easy to access the community contribution guide because it is on every coding challenge's page, but the other guides like the core contribution guide, and content contribution guide are not easy to access from website. I think it would be good to add some way to access all the guides.

violetcraze:
Been working on adding drop down menus to the navigation bar. Open to feedback or suggestions. I'll leave this open for a couple days so people have an opportunity to check it out and make sure it looks good.

Worked off [comment from #1759](https://github.com/CodingTrain/website/pull/1759#issuecomment-544279092) for accessibility. Tested with the following:
- Firefox and Chromium on Windows with [NV Access Screen Reader](https://www.nvaccess.org/)
- Chrome on Android with [Google TalkBack](https://support.google.com/accessibility/android/answer/6283677?hl=en)

fixes #1750 

Here's some screenshots:

navbar1
navbar2
navbar3
navbar4
navbar5
navbar6
navbar7


gruselhaus:
@violetcraze the navbar does not work on Mac using Safari. Could you cross check that?

shiffman:
I can confirm that it's not working on safari, don't seen anything in console.

Also, small thing but the "hand" doesn't appear when hovering the navbar as it does for other interactive elements on the page:

Screen Shot 2019-12-29 at 11 04 21 AM Screen Shot 2019-12-29 at 11 04 31 AM

phucduong86:
It doesn't work on Chrome either on my Mac. Only the Home button works as expected.
Looks like there was some regressions from the latest changes.

vimkat:
Great work, @violetcraze! ^^

It's working for me - but only when you actually click the element. @gruselhaus / @shiffman / @phucduong86 have you tested hovering or clicking? Also adding a `onmouseover` event would make it feel more natural to desktop users.

The "hand" could easily be fixed by applying `cursor: pointer;` to the respected HTML element in CSS. Then it also feels "clicky".

PR #2047, opened 2020-01-10
Link logo to landing page

mikakruschel:
I think it's standard that logos are linked to landing pages. And because on mobile devices the header image of the CTs website is just the logo, I linked it. 
I used the css property `pointer-events: none` so that the header image on desktops don't redirect. 

I'm not sure how this works with screenreaders but we could also add the text "Home" in the `a` element and change the text color to transparent.


gruselhaus:
> I'm not sure how this works with screenreaders but we could also add the text "Home" in the `a` element and change the text color to transparent.

@phucduong86 What are your thoughts?

_jekyll/includes/2-base/header.html
```diff
@@ -4,6 +4,6 @@
 
 
 <header>
-    <div class="header-image"></div>
+    <a href="/" class="header-image"></a>
```
phucduong86: You might want to change this to 
```
<a class="header-image" href="/" aria-label="Home" aria-hidden="true" tabindex="-1"></a>
```
The purposes of them are:
aria-label : just to get around the automated testing that would complain about empty anchor link
aria-hidden and tabindex : so that it is not focusable and removed from the accessibility tree. This way the visually-able mouse user can always click on it to go to the home page, the visually-impaired user won't head the "Home" button twice, and the visually-able-keyboard user don't get confused when they don't know where the mouse focus is.

assets/css/2-base/_header.sass
```diff
@@ -3,6 +3,9 @@ header
 
   .header-image
     height: 30rem
+    display: block
+    pointer-events: none
+    cursor: default
```
phucduong86: I'm not sure if this is necessary.
Might be better to leave this out so that user has a consistent experience across different screen sizes.

mikakruschel: I‘m not sure about that. I wouldn’t expect a header-image to link to the landing page, except the header image is simply the logo.

mikakruschel: And on desktop you can go to the landing page with just one click

mikakruschel: @gruselhaus what do you think about it?

phucduong86: > And on desktop you can go to the landing page with just one click

To be honest I don't think this feature is necessary at all. On mobile, to navigate anywhere else user already have to click on the Menu button anyway, why would the home button be treated differently, and also creating more complexity for user: 
- Inconsistent behavior for the same visual component (the logo image)
- Also create potential issue where user miss click the menu button and get redirected
- If the menu layout is logical, the user shouldn't have to navigate back to the Home page

gruselhaus: I am with @phucduong86

PR #2086, opened 2020-02-07
Changing the language on the cabana page

shiffman:
(Not sure exactly what should go here but this is a little improvement I think.)

### Community Contribution

### Sharing

**Instagram:**

**Twitter:**


violetcraze:
Sounds good we can always change it if we think of something better in the future.

PR #2090, opened 2020-02-08
Fix release dates

mikakruschel:
I used the same script as in #1878 to find wrong release dates.

Especially the release date from the latest "Hilbert Curve" video is off by a year so it won't appear on the landing page.


gruselhaus:
Thank you!

PR #2096, opened 2020-02-09
Contribute a variant of the Hilbert Curve sketch

edorfaus:
### Community Contribution

https://edorfaus.github.io/coding-train-variants/CodingInTheCabana/Cabana_003_Hilbert_Curve/P5/

This variant starts by drawing an order-1 Hilbert Curve, then zooms out and keeps going, until it fills the entire canvas. To make the direction the curve is displayed in consistent, and the zooming easier, the sketch draws the curve in two directions simultaneously (at different speeds), starting from the bottom-left corner of the curve.

The code has some visualization control parameters at the top, but there's currently no UI for changing them.

### Sharing

-- It would be more of a shrug - I guess I don't mind if you really _want_ to share it, but it's not something I particularly care about (since I tend to avoid social media), so it wouldn't make me be happy.


violetcraze:
Nice animation!

PR #2097, opened 2020-02-09
Update overview link for contributing ports

edorfaus:
The old link doesn't work (404), so update it to point to the pinned issue for porting, since that contains an overview table that it says is updated automatically.


violetcraze:
Thanks for the update!

PR #103, opened 2017-03-06
Merge pull request #1 from CodingTrain/master

tslytsly:
merge

PR #2105, opened 2020-02-10
Add port to p5.js of Coding Challenge #103: Fire Effect

edorfaus:
This is a straight port of the current Processing code, making it work with as few changes as I could manage.

The most obvious and biggest differences are the index calculations for pixels arrays, and setting 4 instead of 1 entry at a time; both of which are necessary due to the different structure of these arrays.

I also halved the width/height of the canvas, because that lets it run fast enough to see that something is moving, albeit still slowly.

I can see several easy ways to optimize this code so it runs faster, but I assumed it was more important to stay as close as possible to the original, so it's easier to understand what's going on based on the video.

(Feel free to tell me if I should do/have done something differently here.)

### DIFF
--- a/CodingChallenges/CC_103_Flames/P5/index.html
+++ b/CodingChallenges/CC_103_Flames/P5/index.html
@@ -1,0 +1,16 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+  <meta charset="utf-8" />
+
+  <title>Fire Effect - Coding Challenge #103</title>
+
+  <script src="https://cdn.jsdelivr.net/npm/p5@0.10.2/lib/p5.min.js"></script>
+</head>
+
+<body>
+  <script src="sketch.js"></script>
+</body>
+
+</html>
--- a/CodingChallenges/CC_103_Flames/P5/sketch.js
+++ b/CodingChallenges/CC_103_Flames/P5/sketch.js
@@ -1,0 +1,123 @@
+// Fire Effect
+// The Coding Train / Daniel Shiffman
+// https://thecodingtrain.com/CodingChallenges/103-fire-effect.html
+// https://youtu.be/X0kjv0MozuY
+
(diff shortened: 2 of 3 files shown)


edorfaus:
The first build failed because prettier didn't like some of the formatting from the original code, so I have cleaned it up the way it said to do. Some of those changes make the code a bit less clear in my opinion, so I have also added some comments that hopefully clarify the intent.

violetcraze:
Thanks working on this!! 🔥🔥🔥 Also the comments are great! Feel free to give some advice on how linting with prettier could be better. If it's making the code worse we should look at improving it.

edorfaus:
Well, I don't really have much (if any) advice to give; while the linting setup could probably be improved (as always), I haven't noticed anything in particular I could point to and say "this should be set up like that instead".

The thing is, tools like prettier can't really look at the full context, and changes that improve the code in most places can still make it worse in some other places.

Like in this PR, where it complained about the extra parentheses in `(x + 1) + (y) * w` - usually, such parentheses are extraneous and should be removed, but in this particular case they were there to more clearly separate the X-coordinate from the Y-coordinate. But prettier didn't know that.

An even clearer example I saw elsewhere, is extra spaces used for alignment - e.g. an array of x,y,z coordinate arrays `[ [ 4, 10,   2], [10,  2, 432], ... ]` with each triplet on its own line - the extra spaces (which aren't showing up in the comment preview...) are there to make the coordinates line up into columns, to make the list easier to scan through visually - but in most arrays, argument lists, etc., extra spaces would be a mistake that should be removed. Again, telling the difference requires context that prettier simply doesn't have.

There might be ways to tell prettier about such things, like "ignore that style rule in this section" comments, but littering the code with those is hardly perfect either.

Basically, using tools like prettier means that the code style can never _quite_ be perfect. But on the other hand, _not_ using them is probably worse, because humans make a lot of mistakes... So, yeah. It's a tradeoff.

PR #2122, opened 2020-02-15
Add a port to p5.js of coding challenge 22: Julia Set

edorfaus:
This code started out as a straight translation of the Processing code, just changing the use of the pixels array because it behaves differently. (And added the explanation comment about it.)

Then I added an optimization (with explanation comment), that I think changes the code little enough that it won't really be harder to follow along with - and speeds things up enough that it seemed worth it. It basically brings the sketch up from a frame rate of 0.4 in the web editor, and about 7-8 directly in my browser (I don't know why the difference), to about 50-60 in both. (Though it varies a lot with time as the sketch runs.)

However, I kept that as a separate commit, so if you disagree with me and would prefer the slower but less modified version, we can just remove that commit before merging. (Though in that case, I'd recommend changing the web editor version to use a smaller canvas or something, to speed it up a bit.)

### DIFF
--- a/CodingChallenges/CC_022_JuliaSet/P5/index.html
+++ b/CodingChallenges/CC_022_JuliaSet/P5/index.html
@@ -1,0 +1,16 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+  <meta charset="utf-8" />
+
+  <title>Julia Set - Coding Challenge #22</title>
+
+  <script src="https://cdn.jsdelivr.net/npm/p5@0.10.2/lib/p5.min.js"></script>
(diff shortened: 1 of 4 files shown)


edorfaus:
Ok, fixed some code style issues I missed originally, then squashed that fix into the first commit. I should probably install npm etc. so I can run the linter locally before I push... Oh well.

violetcraze:
Looks great! I think the change is fine with the comments. Hope you don't mind I took the opportunity to add in some other small changes and put the p5 code into the web editor. Thanks for working on this.

PR #2125, opened 2020-02-15
removed the flipped Video after classification

AliKarpuzoglu:
If the video is not removed, it causes new <canvas> elements to be added to the DOM constantly , this could cause a huge spike in memory usage

This is just a quick bugfix


gruselhaus:
Thank you for catching this

PR #2127, opened 2020-02-16
CC 8: Remove unused field in Planet class

edorfaus:
In the videos, this field was not added until challenge 9, and it is not used in the code here.

While I was at it, I also updated the header.


violetcraze:
Thanks for catching this!

PR #2128, opened 2020-02-16
Add port to p5.js of Coding Challenge 8: Solar System part 2 (3D)

edorfaus:
It's as straight a port as I could manage, though some changes were necessary - like using p5.EasyCam instead of PeasyCam. I think it's pretty close overall, though.

Note that this assumes PR #2127 is OK, as it does not include the unused field that is removed by that PR.


violetcraze:
Looks great!! Thanks for working on this p5.EasyCam was a great choice.

PR #2131, opened 2020-02-19
Starting Nature of Code Pages

violetcraze:
Possibly waiting for final video ids.

After merging check to see if repository is linking correctly. I'm not entirely sure on what to put in there but it's impossible to test without the example code in the master.


shiffman:
@violetcraze The final IDs are ready but let's wait to merge / actually put on the website until at least the first one is public.

violetcraze:
@shiffman should be up to date and ready to merge whenever

shiffman:
Working on these pages and will merge shortly, but wanted to just note here that a long description for `index.md` doesn't work well with the design. I think maybe I don't actually need the longer description here anyway but thought I would flag. 

Screen Shot 2020-03-06 at 11 15 21 AM

shiffman:
@violetcraze I removed the pages for 1.1 and 1.2, can we add them back in a separate pull request? (I saved the files just in case)

PR #2132, opened 2020-02-20
Add port to p5.js of CC 9: Solar System Generator part 3 (3D textures).

edorfaus:
Since, at least as far I have been able to find, p5.js does not have a direct equivalent to the `PShape` object, this instead uses the texture image in immediate mode, applying it directly to the sphere that was already being drawn.

Like the Processing version, this is based on the code from CC 8, and has a similar-sized diff, in much the same areas. (At least while ignoring the added comments.)

I wasn't sure about making a copy of the image files from the Processing version, but they're not really all that big, and I figured it would be easier than setting up a symlink, at least on the recipients who might not be familiar with those.


edorfaus:
I'll note that at the time of the previous pull requests (where I removed the not-yet-used `globe` field), I was thinking that I would be adding it back in for the code of this challenge - I hadn't yet realized that `PShape` isn't a thing in p5.

I considered keeping the "globe" name for the texture, but ultimately decided that it would probably be less confusing overall to use a different name, since the field no longer holds the shape of the globe, only the texture image - and the potential confusion of not having the `globe` field anymore seems likely to be smaller (in the long term) than the potential confusion of not immediately noticing that something is significantly different and trying to use the field as a shape.

violetcraze:
Thanks so much for working on this!! Sorry it took a while to merge it.

PR #2137, opened 2020-02-23
fixing links in ml5 pages to the code repos

violetcraze:
Minor fix to repo paths in ml5 pages.


shiffman:
@violetcraze it might be nice to reorganize the directories to mirror the navigation on the website itself, i.e.

```
/learning/ml5/
```

And do away with the "Tutorials" and "Courses" and leftover directories from before?

PR #2144, opened 2020-02-25
Add a port to p5.js of Coding Challenge 58: 3D Earthquake Visualization

edorfaus:
This is as straight a port as I could manage, with the exception of using a smaller data source by default (but still bigger than the tiny one-day source Shiffman used first), and adding a bunch of comments.

This seemed like it would be a pretty simple port before I started working on it, but then I got tripped up by several small problems. (Even ended up starting over once.)

The first was that it threw a `NaN` error somewhere in `draw()`, which turned out to come from `row.getNum('mag')` because one of the rows in the dataset had an empty `mag` field. While I have a fix for that which drops such rows, I eventually worked around it by using a smaller dataset which doesn't include that row, and had the bonus of making the sketch not run as slowly (I think it was doing low single-digit FPS with the full set) - which I ended up using as an excuse to not include the fix, since it would require a bit of explanation despite being just a single line in `setup()`:

```js
  table.rows = table.rows.filter(row => row.getString('mag') !== '');
```

The next thing was mostly my own fault, nothing was being rendered because I forgot to initialize the `angle` variable (which Processing does automatically for floats), and `rotateY` doesn't like being given `undefined`.

The last thing, though, which tripped me up the longest, was that the angles of the boxes were all wrong - while some looked close to sticking straight out, others were nearly flat to the ground. This eventually turned out to be caused by a subtle difference between the `angleBetween` methods in Processing and p5.js, which I only found out about by digging into the source of both and comparing them, and then following the difference there into the PR and issue which added that difference. Basically, the version in p5.js has the additional feature of also returning the direction of the angle, by having a range from -π to π instead of just 0 to π. Luckily, the comments in the PR and issue also gave me a simple solution for what to do when I wanted the Processing version, one which I probably wouldn't have thought of on my own based on just the code.

Anyway, I've tried to explain the changes in the code comments, even one thing that goes directly against what is shown in the video (again due to differences between Processing and p5.js), and I just hope that it ended up being understandable and close enough to the original to be useful.


violetcraze:
🌍 Thanks so much for working on this! Comments are great but you don't always have to explain small differences between p5 and processing. They're only really necessary when someone couldn't easily find it in documentation. Not complaining just seems like you might be doing more work than is needed.

edorfaus:
Yeah, probably - I had a few things I felt needed explaining, like the abs on angleBetween (which isn't obvious from the documentations), and then it just got hard to see where to stop... Oh well. :)

PR #2146, opened 2020-02-29
Add a port to p5.js of Coding Challenge 101: May the 4th Scrolling Text

edorfaus:
This is a straight port, with pretty minimal changes from the original Processing version.

In brief:
- fullscreen removed, since in the browser it requires user interaction
- added a font, since one is required to draw text (it's a copy of the one in challenge 59, so shouldn't present any new copyright issues)
- fixed the calculation of `w` (line 32) to avoid the negative width (Processing apparently uses the normalized area of the box if it's specified backwards like that, but in p5.js the text gets drawn at the given position, which is off to the right side, with just one word per line)


gruselhaus:
Thank you

PR #2149, opened 2020-03-01
Add port to p5.js of Coding Challenge 102: 2D Water Ripple

edorfaus:
This is a straight port - the only real differences (beyond syntax and coding style) are the use of the `pixels` array (index times 4, setting density), and the `current`/`previous` array initialization (which is a bit tricky, not sure if it needs further explanation).

### DIFF
--- a/CodingChallenges/CC_102_WaterRipples/P5/index.html
+++ b/CodingChallenges/CC_102_WaterRipples/P5/index.html
@@ -1,0 +1,16 @@
+<!DOCTYPE html>
+<html>
+
+<head>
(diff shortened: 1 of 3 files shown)


edorfaus:
I noticed there's a PR underway to update the version of p5.js being used. This port currently uses the previous version (0.10.2), but I can update it to the new version (1.0.0) if you want - a quick test showed it works fine with that version too.

violetcraze:
Looks great thanks for working on this! Would be great to update to 1.0.0 so we don't have to go back and do that later.

PR #2151, opened 2020-03-02
Remove an incorrect web editor link.

edorfaus:
It points to the sketch from an earlier coding challenge (CC 138: Angry Bird), and since there's no p5.js version of this challenge (yet), there isn't any correct link to replace it with.


gruselhaus:
Thank you!

PR #2154, opened 2020-03-06
Fix a bug in CC_010_Maze_DFS

tigertv:
It fixes a bug, when a and b are not neighbors.
If a and b are "diagonal" neighbors, then walls still would be broken.

### Community Contribution

https://github.com/tigertv

### Sharing

### DIFF
--- a/CodingChallenges/CC_010_Maze_DFS/P5/sketch.js
+++ b/CodingChallenges/CC_010_Maze_DFS/P5/sketch.js
@@ -68,18 +68,18 @@ function index(i, j) {
 
 function removeWalls(a, b) {
   let x = a.i - b.i;
-  if (x === 1) {
+  let y = a.j - b.j;
+
+  if (x === 1 && y === 0) {
     a.walls[3] = false;
     b.walls[1] = false;
-  } else if (x === -1) {
+  } else if (x === -1 && y === 0) {
     a.walls[1] = false;
     b.walls[3] = false;
-  }
-  let y = a.j - b.j;
-  if (y === 1) {
+  } else if (y === 1 && x === 0) {
     a.walls[0] = false;
     b.walls[2] = false;
-  } else if (y === -1) {
+  } else if (y === -1 && x === 0) {
     a.walls[2] = false;
     b.walls[0] = false;
   }


violetcraze:
@tigertv Appreciate the fix!! Can you reproduce the bug with the current sketch? Because I couldn't. It would definitely be a better function checking both x and y but we like to keep the code from the coding challenges as close to the video as possible so those looking for the code aren't confused when it doesn't match. If you can show it breaking then it would be worth fixing and adding a comment but otherwise I would suggest submitting it as a community contribution. Here's a [guide on community contributions](https://thecodingtrain.com/Guides/community-contribution-guide).

tigertv:
Thanks for your answer. 
OK, @violetcraze. I show you what happens. Let's change the draw function in sketch.js:

```javascript
function draw() {
  background(51);

  for (let i = 0; i < grid.length; i++) {
    grid[i].show();
  }

  let cell1 = grid[index(0, 2)];
  let cell2 = grid[index(1, 0)];
  cell1.visited = true;
  cell2.visited = true;
  removeWalls(cell1, cell2);

  cell1 = grid[index(0, 0)];
  cell2 = grid[index(1, 9)];
  cell1.visited = true;
  cell2.visited = true;
  removeWalls(cell1, cell2);
}
```
When we run it, we get the following result:

image

Here we see that the wall between cells [0, 0] and [1,0] is broken. But the code doesn't use the function `removeWalls` for the cells. So what does it mean? It means that the `removeWalls` doesn't work properly.

violetcraze:
@tigertv Thanks for the explanation.

I agree that that `removeWalls()` could be better but I still don't see how  the problem could happen in using the current draw loop because `current.checkNeighbors()` will only change by one in one direction. I don't see a way that a diagonal could happen without changing some other piece of code in the example. Let me know if I'm wrong though.

If that's the case, I don't think we should change the original code because the main purpose of the examples on the site is to match the code in that was written in the video. While your changes make the code better we'd rather not change something unless it's broken to the point where it doesn't run correctly because people might be confused if they are coming from the video.

Community contributions exist so people can change and improve on code from the videos and share it so that might be a better way to share this improved version.

tigertv:
OK. Thanks.

PR #2181, opened 2020-03-22
Add port to p5.js of Coding Challenge 142.1: Rubiks Cube 1

edorfaus:
This is a straight port, except for using p5.EasyCam instead of PeasyCam, and having to deal with QUADS not being supported in WebGL mode (this issue: processing/p5.js#4401).

I left the workaround for that last bit as a separate commit, to make it easy to revert when that support has been implemented.


shiffman:
Thank you so much for this!!

PR #2302, opened 2020-05-10
fixed potential bug with a tag opening in new tab

GypsyDangerous:
i noticed that `blank` leads to some unexpected actions that are remedied by `_blank` so i think it would be good to change it


gruselhaus:
Thank you @GypsyDangerous

PR #2317, opened 2020-05-12
Adding Support for Multiple Examples

violetcraze:
It seems more and more frequent that we run into problems of not having a good way of sharing multiple sketches / examples for one video. Hopefully this will fix that.

The new feature allows for adding custom variations for each page. Which is done like so:
```
variations:
  - name: "Data Collection Sketch"
    folder: "1_data_collection"
    lang: "p5.js"
    live_example: true
```

This can be in addition to or instead of the normal `P5`, `Processing`, `JavaScript`, and `Node` folders.

I changed the format to allow for more unique names and added the lang tag. These two are combined in the button like `p5.js name`. I know lang isn't correct because processing and p5 are libraries but I couldn't think of a better term.

I also switched all pages using p5-multi to the new system and removed it from the possibilities.

Here's a screenshot of ml5 7.3 with filtering turned off because it checks against the repository. I've highlighted the ones that would be visible on the website.

variation screenshot


shiffman:
This is great, love it! A couple small comments:

* Let's make sure to run a WCAG accessibility test.
* I think we can clean up / simplify the language. Maybe remove "sketch" and "program"? So it's just "p5.js", "Processing", "data collection", etc.
* The one thing that is a little confusing is the use of "web editor". This is the same as the "p5.js" listing but in the web editor rather than github, correct? Maybe it should just always come first? Should we actually just include the web editor and not the github link? So "p5.js" goes to the web editor?

violetcraze:
@shiffman @gruselhaus Sorry it took me so long to following up on this. I made some changes that hopefully fix all the previous problems with the drop downs.

I ended up switching them to badges (for lack of a better word) because the drop downs are difficult to make work with screen readers and even when they are accessible in my opinion they still are tiring to use with a screen reader.

I did a WCAG test and cleaned up naming as well as grouping the web editor with p5.js sketches.

If there's no objections then I'll merge it.

The new variation options look like:
```
variations:
  - name: "Data Collection"
    folder: "1_data_collection"
    lang: "p5.js"
    web_editor: Fe1ZNKw1Z
```

Here's a few screen shots:
variation-examplexcf
variation-colors
variation-moblie

gruselhaus:
Love this new version!!!!

shiffman:
This is fantastic! Let's merge and we can make any further from there! Thank you! 🚀

PR #2372, opened 2020-06-11
Simple Particle System Contribution: Flower Pot

ssmkhrj:
I have created a vibrant and colorful flower pot, which is a very famous cracker that is burnt in the festival of Diwali in India. It is also know as Anaar in Hindi. I have made it look like the sparks (or particles) go up from the source and then they lose their acceleration and come down and eventually disappear.

### Community Contribution

https://codepen.io/ssmkhrj/full/OJMNmJJ

### Sharing

**Instagram:** https://www.instagram.com/ssmkhrj/


gruselhaus:
Great!

PR #2436, opened 2020-07-18
Small fix in cave generation with marching squares

JurriaanD:
This is a 1-character-delta bugfix for the cave generation code from the marching cubes coding challenge.
  
The `countNeighbors` function counts the number of 1's there are in a 3x3 grid centered at the given `(x, y)`. In this function, we iterate over neighbors at indices `(i,j)`. We only want to count real neighbors and not the node itself, so we should check that `(x,y) != (i,j)`. A small typo slipped in causing the function to check that  `(x, x) != (i, j)`.


gruselhaus:
Thank you!

PR #2469, opened 2020-08-03
Mandelbrot Set but in PyGame

AirtonBorges:
The same version of the Mandelbrot Set shown in the video, but made in Python using the PyGame module to visualise it.

### Community Contribution

### Sharing

**Instagram:**

**Twitter:**


AirtonBorges:
I forgot to change the branch name, hope that isn't a problem

PR #2500, opened 2020-08-18
add rel attribute to prevent click-jacking

lpmi-13:
`target="_blank"` is open to click-jacking attacks, so adding `rel="noopener noreferrer"` is a way to prevent that.

As an aside, both `noopener` and `noreferrer` prevent this vulnerability, though `noopener` won't work with some legacy browsers, and newer browsers just ignore `noreferrer`, so adding both is the current accepted best practice.

### DIFF
--- a/404.md
+++ b/404.md
@@ -4,5 +4,5 @@ permalink: /404.html
 ---
 <h2> 404 :( </h2>
 <h2>your rainbow is in another train station</h2>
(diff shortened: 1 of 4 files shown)


lpmi-13:
for reference, here's the current site audit on lighthouse:

click jacking vulnerabilities on current site

and here's the lighthouse score with the current commit:

audit score with current commit

gruselhaus:
Thank you!

PR #2514, opened 2020-09-05
Metaball contribution: individually-colored w/ GLSL

abangfarhan:
Implementation of metaballs written in Processing and GLSL (shader). Each metaball has its own color. See https://www.youtube.com/watch?v=LZfmCuEPRIM for demo.


gruselhaus:
Thank you for this awesome contribution ❤️

PR #2667, opened 2020-10-05
created web editor version of RiveScript chatbot

shiffman:
Also updated the code to work with latest RiveScript.


violetcraze:
I'll check to see if there's other example code and if so I'll update it with another pr.

PR #2670, opened 2020-10-06
Adding back jest and seperating linting in actions

violetcraze:


violetcraze:
Decided to bring back build test because while feature peek is helpful it's much slower than a normal build test. So I think we'll make it an optional test that is not required before merging pull requests.

PR #2807, opened 2020-10-25
Seven-segment contribution: editable appearance

NCProductions:
You can change the width and the height of each segment and the space between the segments too, thanks to sliders under the canvas.
Also, there's a little animation in the beginning (a function that allows you to animate the value that's displayed a given number of times).
Link to demo: https://editor.p5js.org/NC_Productions/full/_bNJgU6kF


violetcraze (approved):
Looks great! Thanks for sharing.

PR #2822, opened 2020-10-29
Update 157-zoom-annotations.md

shiffman:
I'm not sure to what extent this really matters but updating the date. It's members only Oct 28 and will be public around 9am ET Oct 29.


violetcraze:
As far as I can tell the date is so streams don't show up before they happen. So I often will put a date that was in the past so I can see the page when doing local builds but often forget to update it later on.

PR #2838, opened 2020-10-31
Update 157-zoom-annotations.md

ItayNaveh:
I would be happy if this was shared on the Coding Train channel but not on other social media


violetcraze (approved):
Thanks for sharing!

violetcraze:
Sorry having trouble merging will have to look into why it's not working later. It's nothing on your end looks like it might be a github bug.

Issue #2762, opened 2020-10-20
Add Missing Pages in Programing with Text Series 💫


violetcraze:
A lot of the Programming with Text videos are missing pages we could you're help adding these pages. 🌠

If you'd like to work on adding a page. Here's the steps.

1. 🌼 **Choose a video from the list and comment to claim it.** Below there's a list of videos that need pages to be created. Comment on this issue to claim one so people aren't working on the same one. *Be sure to check and see if there's recent comments because checkboxes on the list might not be entirely up to date.*

> - *Note: Only take on 1-3 videos at a time, but feel free take on more after you've submitted a pull request.*

2. 🌺 **Follow the [Pages Contribution Guide](https://thecodingtrain.com/Guides/pages-contribution-guide.html) to add a page for the video.** In this case the pages should go under `_learning/programing-with-text` and any code should go in `learning/programing-with-text`. If you're confused about any of the steps in the guide feel free to comment with questions on this issue.

> - Code maybe located: https://github.com/shiffman/A2Z-F16 or https://github.com/CodingTrain/website/tree/main/Courses/programming_with_text
> - Pages Contribution Guide Link: https://thecodingtrain.com/Guides/pages-contribution-guide.html

3. 🌸 **Make a pull request with the new page!!**

> - *Note: Please don't say your pull request closes this issue. I'll close it when all the pages are done.*

**List of Programing with Text Videos**

|Claimed|Done|Number|Link|
|-|-|-|-|
||<ul><li> - [X] </li></ul>| 1.1 | https://youtu.be/HRANU6KtNEs |
||<ul><li> - [X] </li></ul>| 1.2 | https://youtu.be/d3OcFexe9Ik |
||<ul><li> - [X] </li></ul>| 1.3 | https://youtu.be/uNQSVU0IKec |
||<ul><li> - [X] </li></ul>| 1.4 | https://youtu.be/0Mq2CxspF5s |
||<ul><li> - [X] </li></ul>| 1.5 | https://youtu.be/DcoAjEZYies |
||<ul><li> - [X] </li></ul>| 1.6 | https://youtu.be/YTzVbuV89RE |
||<ul><li> - [X] </li></ul>| 2.1 | https://youtu.be/7DG3kCDx53c |
||<ul><li> - [X] </li></ul>| 2.2 | https://youtu.be/YTocEnDsMNw |
||<ul><li> - [X] </li></ul>| 2.3 | https://youtu.be/EfJU0Y9WAZ4 |
[...]

PR #2846, opened 2020-11-02
Added md files programming-with-text video 9.1-9.3

KobeLiesenborgs:
Add "md files programming-with-text video 9.1-9.3" for issue #2762


violetcraze (approved):
Looks great! Thanks for working on this.

Side note: I think @gruselhaus might have found the reason you were having trouble. Looks like the `Gemfile.lock` might have gotten updated automatically or something. I seem to remember having a similar problem at some point. Let me know if that was the problem.

Issue #2876, opened 2020-11-13
Remove outlines


harshhhdev:
![image](https://user-images.githubusercontent.com/69592270/99012944-71e45200-251d-11eb-800e-c67bd07cef65.png)

Right now, when you click on a button there is a weird outline on it. This may be intentional, but in my opinon it doesn't look that good and feels weird/out of place.

violetcraze:
I believe this might be for accessibility. I'll have to look into it more. It'll do my best to do that in the next 3 days or so.

Also would like to hear @phucduong86's opinion.

If it turns out to not be important to accessibility I wouldn't be against removing it.

harshhhdev:
A bit curious, not trying to bicker or anything.

How're these outlines helping accessibility? I usually have a tendency to take them out in my personal projects and I was wondering if I should stop doing that.

calafm:
Hello! Since I didn't want to commit directly, I made a pull request on this. 

I removed the white-border from the span. Hope it's alright. :)

phucduong86:
Hi,

Sorry I haven't had a chance to respond to this earlier.
It is required for accessibility so that user know it's the focused element.

harshhhdev:
Oh, ok. Thanks.

PR #2886, opened 2020-11-18
Deleting the span's white border in navigation menu

calafm:
I believe the website would look better without the white border on :focus in navigation-menu.
Edit: Ignore the Readme, I deleted a letter by mistake but I put it back.


violetcraze:
Thanks for this contribution however it was determined for that the outline is needed for accessibility in #2876 so won't be merging this.

PR #3008, opened 2021-01-21
Removing Unused Variable

arnavsirigere:
Removed the unused variable `nn`


gruselhaus:
Good catch! Thank you!

Issue #3022, opened 2021-02-03
htmlproofer fails currently in CI


gruselhaus:
```
2021-02-03T17:48:26.9706473Z ##[group]Run bundle exec htmlproofer ./_site --assume-extension --check-html --disable-external --url-ignore "/#.*/"
2021-02-03T17:48:26.9707667Z [36;1mbundle exec htmlproofer ./_site --assume-extension --check-html --disable-external --url-ignore "/#.*/"[0m
2021-02-03T17:48:26.9758004Z shell: /usr/bin/bash -e {0}
2021-02-03T17:48:26.9758441Z ##[endgroup]
2021-02-03T17:48:27.4752941Z LoadError: libffi.so.6: cannot open shared object file: No such file or directory - /home/runner/work/website/website/vendor/bundle/ruby/2.6.0/gems/ffi-1.14.2/lib/ffi_c.so
2021-02-03T17:48:27.4754344Z   /home/runner/work/website/website/vendor/bundle/ruby/2.6.0/gems/ffi-1.14.2/lib/ffi.rb:6:in `require'
2021-02-03T17:48:27.4755487Z   /home/runner/work/website/website/vendor/bundle/ruby/2.6.0/gems/ffi-1.14.2/lib/ffi.rb:6:in `rescue in <top (required)>'
2021-02-03T17:48:27.4756613Z   /home/runner/work/website/website/vendor/bundle/ruby/2.6.0/gems/ffi-1.14.2/lib/ffi.rb:3:in `<top (required)>'
2021-02-03T17:48:27.4757876Z   /home/runner/work/website/website/vendor/bundle/ruby/2.6.0/gems/ethon-0.12.0/lib/ethon.rb:2:in `require'
2021-02-03T17:48:27.4759059Z   /home/runner/work/website/website/vendor/bundle/ruby/2.6.0/gems/ethon-0.12.0/lib/ethon.rb:2:in `<top (required)>'
2021-02-03T17:48:27.4760231Z   /home/runner/work/website/website/vendor/bundle/ruby/2.6.0/gems/typhoeus-1.4.0/lib/typhoeus.rb:2:in `require'
2021-02-03T17:48:27.4761591Z   /home/runner/work/website/website/vendor/bundle/ruby/2.6.0/gems/typhoeus-1.4.0/lib/typhoeus.rb:2:in `<top (required)>'
[...]

PR #3023, opened 2021-02-03
Testing for #3022

gruselhaus:
This PR is invalid. I'm making sure that the htmlproofer fails do not relate to some dependency caching issues


gruselhaus:
Yea, looks like a chaching issue

gruselhaus:
#3022 was resolved

PR #3045, opened 2021-02-12
CodingChallenges / Fluid Simulation / P5 (function renderV fix)

LeithBA:
### Description

This was probably not noticed before because the main loop does not use this function at all.

-removed leftover 3rd dimention variable "k" from line 88 and 89
-removed leftover "this.canvas." from line 90

### Community Contribution

[You can try the fix here](https://editor.p5js.org/LeithBA/sketches/aJuYw0KAs)

### Sharing

- [X] I would be happy to have my project shared on The Coding Train's social media!

**Instagram:**
@leithba
**Twitter:**
@leithba


gruselhaus:
This makes sense to me. What do you think @shiffman ?

LeithBA:
I think this was forgotten 😅

KobeLiesenborgs:
@LeithBA it was indeed! Merged!

PR #3061, opened 2021-02-16
Port CC_037_diastic to Processing

spencerstith:
### Community Contribution

CC_037_diastic working

### Sharing


spencerstith:
I will say that I had to make some modifications for this to work since the original was made with P5.js and Processing doesn't have anything like the DOM library or getting input from the console. This was done by just having the user put their seed into the sketch window and utilizing the keyPressed() function.

shiffman (approved):
This is wonderful, thank you!

PR #3065, opened 2021-02-20
Pendulum Variations

shiffman:
Adding these variation for the pendulum! @violetcraze I'll merge this if it passes tests but I may have missed some details so feel free to review and let me know if I did! (I didn't add the code to local repo folders for example).


shiffman:
I'm releasing this video so am going to merge this even with the test failing, the "FeaturePeek" looks great to me. @violetcraze if you could take a look when you have a chance if I messed something up?

gruselhaus:
Found the reason why the test failed. The property `folder` of variation is currently mandatory.

PR #3079, opened 2021-02-24
Removing the Patreon reference from the supporting section

gruselhaus:
I think we don't use Patreon anymore.


violetcraze (approved):
Looks good to me

PR #3090, opened 2021-03-02
Remove arguments from constructor in sketch.js

adamantonio:
Constructor does not use arguments in this sketch and the concept behind them is not introduced until the next video


gruselhaus (approved):
LGTM

PR #3097, opened 2021-03-08
Moving old p5 2.3 2.4 and 2.6 to archive

duskvirkus:
Relates to #3095 and #3096


shiffman:
We can move `map()` for now from 2.4-archive to the playlist at 2.5. I may redo it but for now the old version can stand but be renumbered.

shiffman:
Ah, also 2.6 can stay as well. So the new sequence is:

* 2.1 - NEW
* 2.2 - NEW
* 2.3 - NEW
* 2.4 - NEW random()
* 2.5 - OLD 2.4 map() renumbered
* 2.6 - STAYS THE SAME

Should I hold off on merging this until we conform to the above? Or merge now and adjust after the fact?

shiffman:
Ah, and we will also have to change and existing references to these videos. For example:

Screen Shot 2021-03-08 at 10 49 36 AM

PR #3157, opened 2021-04-12
Fix bug for base 10 numbers

### LINKED ISSUE #3114: You may like to fix this...
Hey Dan! Thanks for your great video. This is not very serious, but you may change the following line:

https://github.com/CodingTrain/website/blob/d56dbfe30d86a1374a64742108abaf09c5405a97/CodingChallenges/CC_161_pi_from_random/from-random/sketch.js#L25

to
```javascript
randomSequence.push(parseInt(num, 10));
```
as your numbers may start with a `0` and if there is no digit above `7`, it will be interpreted in base `8` by JS.

Good work, though!

ameybhavsar24:
Fixes #3114


gruselhaus:
LGTM

gruselhaus (approved):
LGTM

duskvirkus:
Added a comment so people aren't confused about why the code looks different. Otherwise this is great. @ameybhavsar24 Thanks for the fix!

PR #3309, opened 2021-08-19
Adding page for cc163

duskvirkus:
The normal deal if description looks good then it should be good to merge.

Tried something a bit different with the title. Veratasium recently made a video about how intriguing titles and thumbnails lead to reaching and educating more people. Was somewhat convinced by his argument. If you don't like it then I'm happy to change it back to the normal format.


shiffman:
yes, I saw this video too! Let's try it! I might tweak it a bit.

PR #3436, opened 2021-11-26
Constrain index value

ixley:
Constrain index value for grid[] to avoid `ArrayIndexOutOfBoundsException` error. This seems like an inelegant fix but functionally keeps the index value from exceeding the expected range and throwing an error. If there is a more appropriate way to solve this, I'd love to see it!


KobeLiesenborgs:
Keeping sourcecode in line with the corresponding video

PR #3646, opened 2022-06-01
ulam hexagon spiral variant in python

PWilson-1978:
I have a 3 radii hexagon and all of the primes line up on the radii associated with 5 and 7.  I have another version that opens a file of primes up to 5000 if the square root of a number is larger, it will fill in the primes between.   I modified the routine that checks for the prime and added a check for the last digit of 5 in the check with division of 2 because any number that has a 5 will be divisible by 5, and then I'm also testing another qualification if the modulo of the value being checked is 1 or 5, it may be a prime. Both print out the lists of prime values at the conclusion of the run.

### Community Contribution

### Sharing

**Instagram:**

**Twitter:**


PWilson-1978:
Here's my youtube video of a run.[
](https://youtu.be/7thwHM0Yj_U)

